PR #1, opened 2020-06-08
Codifies flask api helper example

urimandujano:
Adds example for the Flask API helpers which can be run, linted, and tested.  This should help us further verify that changes made to the toolkit preserve functionality.  Breaking changes should be caught before a release is made.  Additionally, the example demonstrates how one can test an ActionProvider.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -903,4 +903,5 @@ also provide validation of input ``ActionRequest`` data to the ``/run`` method
 prior to invoking the ``action_run`` function. As long as the return value from
 the various functions is of type ``ActionStatus``, the framework will also
 insure that the returned JSON data conforms to the Action Provider Interface.
-The example in the Appendix demonstrates how these functions can be implemented.
+The **watchasay** example in the ``examples/`` directory demonstrates how these
+functions can be implemented.
--- a/README.rst
+++ b/README.rst
@@ -491,7 +491,7 @@ Server and create required Scopes.
             library will not be able to, and will therefore not attempt to, retrieve
             a user's groups and so no policies based on Groups may be used. We
             encourage you to consult the `Globus Auth Documentation
-            <https://docs.globus.org/api/auth/>`_ for more information on creation
+            <https://docs.globus.org/api/auth/>`__ for more information on creation
             and management of Scopes for more advanced scenarios such as other
             dependent Globus Auth based services such as Globus Transfer.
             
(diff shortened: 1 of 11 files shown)


README.rst
```diff
@@ -491,7 +491,7 @@ Server and create required Scopes.
             library will not be able to, and will therefore not attempt to, retrieve
             a user's groups and so no policies based on Groups may be used. We
             encourage you to consult the `Globus Auth Documentation
-            <https://docs.globus.org/api/auth/>`_ for more information on creation
+            <https://docs.globus.org/api/auth/>`__ for more information on creation
```
jimPruyne, line 494: Should this be a double underscore? Perhaps that's RST I don't know off the top of my head.

README.rst
```diff
@@ -901,189 +903,4 @@ also provide validation of input ``ActionRequest`` data to the ``/run`` method
...
-            action_release,
-        )
-        app.register_blueprint(skeleton_blueprint)
-        app.run(debug=True)
-
-
-    if __name__ == "__main__":
-        main()
-
+The example in the Appendix demonstrates how these functions can be implemented.
```
jimPruyne, line 906: There's no appendix any more right?

examples/watchasay/app/schema.json
```diff
@@ -0,0 +1,15 @@
+{
+    "$id": "https://automate.globus.org/skeleton_action_provider.input.schema.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Skeleton Action Provider Input Schema",
+    "type": "object",
+    "properties": {
+        "input_string": {
```
jimPruyne, line 7: Maybe good to add a description property here more or less to set the example that it is good to document your properties this way.

README.rst
urimandujano: Yea the rst linter I'm running caught that as a `Duplicate explicit target name error`. Apparently, using a double underscore marks the link as an anonymous reference (seen from this [stack overflow q](https://github.com/sphinx-doc/sphinx/issues/3921#issuecomment-315581557))

README.rst
urimandujano: Yup, good catch!

examples/watchasay/app/schema.json
urimandujano: This would be good too -- I didn't even know the description keyword was an accepted property.  I expect whatever we document here will get copied around so adding it will make a difference.

urimandujano: I updated both examples' schema's property definitions with the description

README.rst
jimPruyne: that's new to me. What rst linter are you using? Perhaps I should do it also (and maybe we should make it a test target of some sort).

PR #3, opened 2020-07-07
Adds introduction and details on Action Provider Interface

urimandujano:
- Uses a readthedocs material theme
- Uses interactive terminal session graphics
- Adds raw AP examples to site
- Adds license page
- Adds installation page


urimandujano:
@jimPruyne since this PR is mostly documentation, you can see a live version of this branch's changes at https://action-provider-tools.readthedocs.io/en/enhancement-init-readthedocs/index.html - it'll probably be easier to see the work visually.

jimPruyne:
This is looking quite good. Style/layout wise I'll say I like the use of the interactive terminal. I'm not a huge fan of the source code color theme, but it doesn't really matter.

In terms of the content, it seems to be looking pretty good. I think the main thing I see missing is something on the examples base page giving some outline of what the examples are. One using the core toolkit, the other using the flask helper. Probably not much more than that is needed.

Then, I'd say we consider this first release of these docs on readthedocs.

urimandujano:
[ch2127]

PR #8, opened 2020-09-09
Adds optional enumeration endpoint to Flask helpers

urimandujano:
- Adds new, hidden API routes to both Flask helpers
- By default, Flask apps created using the Flask helpers can now "register" enumeration endpoints
- Adds documentation on the enumeration endpoint to readthedocs
- Adds tests validating "old style" and "new style" API routes for the Flask helpers are properly setup
- Adds tests to the Flask decorator helper example app and adds these tests to CI pipeline
- Adds example enumeration endpoint implementation to examples
- Adds tests validating enumeration endpoint is accessible on the examples
- Fixes an issue in the patching library where a string was getting parsed as a set
- Adds a patch to the testing library to mock the TokenChecker in the Flask decorators helper
[ch3172]

### DIFF
--- a/examples/apt_blueprint/tests/test_action_provider.py
+++ b/examples/apt_blueprint/tests/test_action_provider.py
@@ -1,4 +1,3 @@
-import json
 import uuid
 
 import pytest
@@ -21,16 +20,25 @@ def client():
         yield app.test_client()
 
 
+@pytest.fixture(scope="function")
+def running_action_id(client):
+    data = {
+        "request_id": str(uuid.uuid4()),
+        "body": {"utc_offset": "1"},
+    }
+    response = client.post("/apt/run", json=data)
+    return response.json["action_id"]
+
+
 def test_introspection_endpoint(client):
     response = client.get("/apt")
 
     assert response.status_code == 200
-    assert json.loads(response.data)["input_schema"] == description.input_schema
+    assert response.json["input_schema"] == description.input_schema
 
 
 def test_enumeration_endpoint(client):
     response = client.get("/apt/actions")
-    print(response.data)
     assert response.status_code == 200
 
 
@@ -40,53 +48,30 @@ def test_run_endpoint(client):
         "request_id": request_id,
         "body": {"utc_offset": "1"},
     }
-    response = client.post("/apt/run", data=json.dumps(data))
+    response = client.post("/apt/run", json=data)
 
     assert response.status_code == 201
-    assert json.loads(response.data)["status"] == ActionStatusValue.ACTIVE.name
+    assert response.json["status"] == ActionStatusValue.ACTIVE.name
 
 
-def test_status_endpoint(client):
-    request_id = str(uuid.uuid4())
-    data = {
-        "request_id": request_id,
-        "body": {"utc_offset": "1"},
-    }
-    response = client.post("/apt/run", data=json.dumps(data))
-    action_id = json.loads(response.data)["action_id"]
-    response = client.get(f"/apt/{action_id}/status")
+def test_status_endpoint(client, running_action_id):
+    response = client.get(f"/apt/{running_action_id}/status")
 
     assert response.status_code == 200
-    assert json.loads(response.data)["status"] == ActionStatusValue.ACTIVE.name
+    assert response.json["status"] == ActionStatusValue.ACTIVE.name
 
 
-def test_cancel_endpoint(client):
-    request_id = str(uuid.uuid4())
-    data = {
-        "request_id": request_id,
-        "body": {"utc_offset": "1"},
-    }
-    response = client.post("/apt/run", data=json.dumps(data))
-    action_id = json.loads(response.data)["action_id"]
-
-    response = client.post(f"/apt/{action_id}/cancel")
+def test_cancel_endpoint(client, running_action_id):
+    response = client.post(f"/apt/{running_action_id}/cancel")
 
     assert response.status_code == 200
-    assert json.loads(response.data)["status"] == ActionStatusValue.FAILED.name
+    assert response.json["status"] == ActionStatusValue.FAILED.name
 
 
-def test_release_endpoint(client):
-    request_id = str(uuid.uuid4())
-    data = {
-        "request_id": request_id,
-        "body": {"utc_offset": "1"},
-    }
-    response = client.post("/apt/run", data=json.dumps(data))
-    action_id = json.loads(response.data)["action_id"]
-
+def test_release_endpoint(client, running_action_id):
     # Cancel before releasing since our action is asynchronous
-    response = client.post(f"/apt/{action_id}/cancel")
-    response = client.post(f"/apt/{action_id}/release")
+    response = client.post(f"/apt/{running_action_id}/cancel")
+    response = client.post(f"/apt/{running_action_id}/release")
 
     assert response.status_code == 200
-    assert json.loads(response.data)["status"] == ActionStatusValue.FAILED.name
+    assert response.json["status"] == ActionStatusValue.FAILED.name
--- a/examples/apt_blueprint/tests/test_action_provider.py
+++ b/examples/apt_blueprint/tests/test_action_provider.py
@@ -1,0 +1,92 @@
+import json
+import uuid
+
+import pytest
+
+from examples.apt_blueprint.app import create_app
+from examples.apt_blueprint.blueprint import description
+from globus_action_provider_tools.data_types import ActionStatusValue
+from globus_action_provider_tools.testing.patches import (
+    flask_blueprint_tokenchecker_patch,
+)
+
+
+@pytest.fixture(scope="module")
+def client():
+    # Create the app in a patched context so that the Provider can startup
+    # without valid credentials AND requests can be made without supplying a
+    # valid token
+    with flask_blueprint_tokenchecker_patch:
+        app = create_app()
+        yield app.test_client()
+
+
+def test_introspection_endpoint(client):
+    response = client.get("/apt")
+
+    assert response.status_code == 200
+    assert json.loads(response.data)["input_schema"] == description.input_schema
+
+
+def test_enumeration_endpoint(client):
+    response = client.get("/apt/actions")
+    print(response.data)
+    assert response.status_code == 200
+
+
+def test_run_endpoint(client):
+    request_id = str(uuid.uuid4())
+    data = {
+        "request_id": request_id,
+        "body": {"utc_offset": "1"},
+    }
+    response = client.post("/apt/run", data=json.dumps(data))
+
+    assert response.status_code == 201
+    assert json.loads(response.data)["status"] == ActionStatusValue.ACTIVE.name
+
+
+def test_status_endpoint(client):
+    request_id = str(uuid.uuid4())
+    data = {
+        "request_id": request_id,
+        "body": {"utc_offset": "1"},
+    }
+    response = client.post("/apt/run", data=json.dumps(data))
+    action_id = json.loads(response.data)["action_id"]
+    response = client.get(f"/apt/{action_id}/status")
+
+    assert response.status_code == 200
+    assert json.loads(response.data)["status"] == ActionStatusValue.ACTIVE.name
+
+
+def test_cancel_endpoint(client):
+    request_id = str(uuid.uuid4())
+    data = {
+        "request_id": request_id,
+        "body": {"utc_offset": "1"},
+    }
+    response = client.post("/apt/run", data=json.dumps(data))
+    action_id = json.loads(response.data)["action_id"]
+
+    response = client.post(f"/apt/{action_id}/cancel")
+
+    assert response.status_code == 200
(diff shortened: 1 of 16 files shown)


examples/watchasay/tests/test_action_provider.py
```diff
@@ -47,6 +47,11 @@ def test_introspection_endpoint(client):
     assert json.loads(response.data)["input_schema"] == load_schema()
 
 
+def test_enumeration_endpoint(client):
+    response = client.get("/skeleton/actions")
+    assert response.status_code == 200
```
jimPruyne, line 52: can we assert anything about the return content? size (even if 0)?

globus_action_provider_tools/flask/api_helpers.py
```diff
@@ -257,6 +264,7 @@ def action_introspect() -> ViewReturn:
             raise NotFound()
         return jsonify(asdict(provider_description)), 200
 
+    @blueprint.route("/actions", methods=["POST"])
```
jimPruyne, line 267: I still think I like this, but I still want to keep it undocumented for now.

globus_action_provider_tools/flask/api_helpers.py
```diff
@@ -288,20 +296,23 @@ def action_run() -> ViewReturn:
...
     @blueprint.route("/<string:action_id>/cancel", methods=["POST"])
+    @blueprint.route("/actions/<string:action_id>/cancel", methods=["POST"])
     def action_cancel(action_id: str) -> ViewReturn:
         auth_state = _check_token(request, checker)
         return _action_status_return_to_view_return(
             action_cancel_callback(action_id, auth_state), 200
         )
 
     @blueprint.route("/<string:action_id>/release", methods=["POST"])
+    @blueprint.route("/actions/<string:action_id>/release", methods=["POST"])
```
jimPruyne, line 315: I think this could just be `DELETE /actions/<string:action_id>`

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -153,6 +179,47 @@ def action_introspect(self):
...
 
+    def action_enumerate(self, func: ActionEnumerationType):
+        """
+        Decorates a function to be run as an Action Provider's enumeration
+        endpoint.
+        """
+
+        @functools.wraps(func)
+        def wrapper():
+            if not g.auth_state.check_authorization(
```
jimPruyne, line 190: Can we somehow make use of the implementation in flask_helpers to avoid some repetition? Maybe that requires a bit of re-factor to allow. Maybe not worth it if we ultimately deprecate the non-Blueprint registration method though.

globus_action_provider_tools/flask/helpers.py
```diff
@@ -0,0 +1,48 @@
+from enum import Enum
+from typing import Iterable, List, Set
+
+from flask import request
+
+
+def parse_query_args(
```
jimPruyne, line 7: can we make the `flask.request` object a parameter here so that we don't have so much reliance on the global-ish state? Perhaps allows for some weird future where we have a request object that's been created by some other means than the incoming call on a flask route.

globus_action_provider_tools/flask/helpers.py
```diff
@@ -0,0 +1,48 @@
...
+    arg_name: str, default_value: str = "", valid_vals: List[str] = None
```
jimPruyne, line 8: `valid_vals` as a `Set` seems to make sense esp since this is returning a `Set`.

globus_action_provider_tools/flask/helpers.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # A query param name with no value returns empty string even with default value
+    if param_val == "":
+        param_val = default_value
+
+    # Split in case there's a comma seperated query param value
+    param_vals = set(param_val.split(","))
+
+    # Remove invalid data from query params
+    if valid_vals is not None:
```
jimPruyne, line 28: If it weren't for the `lower` thing, which is probably a good idea, this could probably just be an intersect operation or some such. That'd be pretty elegant...

globus_action_provider_tools/flask/helpers.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def query_args_to_enum(args: Iterable[str], enum_class: Enum):
+    """
+    Helper function to return an arg value as a valid value for an Enum
+    """
+    new_args = set()
+    for arg in args:
+        try:
+            new_arg = enum_class[arg.upper()]  # type: ignore
```
jimPruyne, line 44: for generality, might not want to always assume `upper`. Perhaps try the provided val and if that fails then try again with `upper`?

globus_action_provider_tools/testing/mocks.py
```diff
@@ -39,7 +39,7 @@ def mock_introspect_token(*args, **kwargs):
     patch(
         "globus_action_provider_tools.authentication.AuthState.identities",
         new_callable=PropertyMock,
-        return_value=frozenset("MOCK_USER"),
+        return_value=frozenset(["MOCK_USER"]),
```
jimPruyne, line 42: oops, that seems like a good catch. Did this end up as a set of `{'M', 'O', 'C', ...}` otherwise. Hate those strings are iterable errors.

globus_action_provider_tools/testing/patches.py
```diff
@@ -6,3 +6,8 @@
     "globus_action_provider_tools.flask.api_helpers.TokenChecker.check_token",
     return_value=mock_authstate(),
 )
+
+flask_blueprint_tokenchecker_patch = patch(
```
jimPruyne, line 10: I need to make sure I'm up to speed on proper use of these mocks etc. I want to be sure I'm using them properly and as much as possible in any AP test cases I've created.

jimPruyne (commented):
Finally got back to this. Sorry it took so long.

Probably a bunch of stuff to be discussed further. In general, I like the new method, and I like the shadow URLs for now.

WRT the shadow URLs, I think we're probably due for a team meeting to discuss future/evolution of our APIs. This is good experience to bring into those conversations.

We'll have to decide if we want to keep the init routes registration in addition to the blueprint. Upon reading this (again) I see the redundancy in the code due to this.

I also admit to glazing over a bit on the details of tests. It continue to be an area where I wonder what more we can do to help users.Perhaps that's mostly a matter of documenting it to users or for me to simply look at existing docs and examples outside the context of diffs which is where I seem to be reading most of this code recently.

examples/apt_blueprint/blueprint.py
```diff
@@ -54,6 +56,49 @@
...
+            "roles": {"creator_id"}
+        }
+
+    Notice that the value for the "statuses" key is an Enum value.
+    """
+    statuses = params["statuses"]
+    roles = params["roles"]
+    matches = []
+
+    for _, action in simple_backend.items():
```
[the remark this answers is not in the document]
urimandujano: Unfortunately the entire implementation is specific to this example and pretty unrealistic (since it's using a dict as the data store). As I was thinking through implementing enumeration for the example, I realized iterating through every action would, in practice, be rough. So this endpoint would require smart use of indexes/filters to offload as much work as possible from python to the storage backend.

Given how different backend implementations can be, all I can think of is defining an abstract base class and asking users to implement their own subclasses. We could use some of the `dynamo_helpers` + experiences using the helpers to define an ABC interface. I'm also open to discussing other ideas

examples/apt_blueprint/tests/test_action_provider.py
```diff
@@ -0,0 +1,92 @@
...
+from examples.apt_blueprint.app import create_app
+from examples.apt_blueprint.blueprint import description
+from globus_action_provider_tools.data_types import ActionStatusValue
+from globus_action_provider_tools.testing.patches import (
+    flask_blueprint_tokenchecker_patch,
+)
+
+
+@pytest.fixture(scope="module")
+def client():
```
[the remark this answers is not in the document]
urimandujano: Some of this is in the toolkit, but is in a haphazard state. As it exists currently, there are some problems with isolating  test cases since the patches patch the base class entirely, and if there are simultaneous tests running that expect a non-patched TokenChecker there'll be issues. I'm looking to patch(heh) the test helpers so they can be used a little more freely and add documentation to using them.

examples/apt_blueprint/tests/test_action_provider.py
```diff
@@ -0,0 +1,92 @@
...
+def test_run_endpoint(client):
+    request_id = str(uuid.uuid4())
+    data = {
+        "request_id": request_id,
+        "body": {"utc_offset": "1"},
+    }
+    response = client.post("/apt/run", data=json.dumps(data))
+
+    assert response.status_code == 201
+    assert json.loads(response.data)["status"] == ActionStatusValue.ACTIVE.name
```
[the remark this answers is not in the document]
urimandujano: Cool I did not know that was a built in --- that is really helpful going to make the changes throughout

globus_action_provider_tools/flask/api_helpers.py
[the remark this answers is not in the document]
urimandujano: I thought about making it a `DELETE` and can't decide why that didn't make it in. I like that method a little more, and I think the route will end up being `DELETE actions/<string:action_id>`

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -153,6 +179,47 @@ def action_introspect(self):
...
+            enumeration = func(g.auth_state, query_params)
+            return jsonify(enumeration), 200
+
+        self.add_url_rule(
+            "/actions",
+            "action_enumerate",
+            wrapper,
+            methods=["GET"],
+        )
+        print(f'Registered action enumeration plugin "{func.__name__}"')
```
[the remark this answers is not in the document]
urimandujano: I had a bunch of these sprinkled throughout to validate that routes were getting registered correctly. I thought leaving them would be a good way of providing feedback as to which routes get registered but I don't think they do that in their current form (after seeing them used in Automate). Removing until we add some logging.

examples/watchasay/tests/test_action_provider.py
urimandujano: I made a change where there's an action run in this test so that enumeration actually returns something. It also helps validate that the query strings are case insensitive.

globus_action_provider_tools/testing/mocks.py
urimandujano: Exactly, I'm sad it masqueraded around as long as it did to begin with...

globus_action_provider_tools/testing/patches.py
urimandujano: When I get a chance to revisit I'll add a page in the documentation for how to use the test helpers. There's an example of using them in each of the Flask helper sample apps but formalizing it with a description would be better for sure.

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -153,6 +179,47 @@ def action_introspect(self):
...
+                self.provider_description.runnable_by,
+                allow_public=True,
+                allow_all_authenticated_users=True,
+            ):
+                current_app.logger.info(
+                    f'User "{g.auth_state.effective_identity}" is unauthorized to enumerate Actions'
+                )
+                raise NotFound
+
+            valid_statuses = set(e.name.lower() for e in ActionStatusValue)
```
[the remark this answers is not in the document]
urimandujano: I switched over to using casefold() just because it's name sorta implies that it's trying to do some case-insenstive comparisons.  The case insenstive checks aren't perfect but Enums names are kinda annoying to work with.

globus_action_provider_tools/flask/helpers.py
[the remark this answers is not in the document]
urimandujano: I've seen APIs use something like "all" as a query param but * might work too. Were you thinking of getting some functionality such is if the value of `arg_name` is "*"  then every `str` in `valid_vals` is returned as part of the parsed result?

urimandujano:
@jimPruyne  thanks a lot for the exhaustive review! I made a bunch of changes which ended up really slimming down the test suite. A takeway that I want to address at some point is documenting the testing strategy with a document walking through the different tools and what they're doing and how to use them.

Mind taking another look at comments/updates?

globus_action_provider_tools/flask/api_helpers.py
```diff
@@ -333,13 +333,17 @@ def action_log(action_id: str) -> ViewReturn:
         def action_enumeration():
             auth_state = _check_token(request, checker)
 
-            valid_statuses = set(e.name.lower() for e in ActionStatusValue)
+            valid_statuses = set(e.name.casefold() for e in ActionStatusValue)
```
jimPruyne, line 336: :+1:

jimPruyne (approved):
I think this looks good now. There's a lot of good evolution of the code base here. We probably won't publicize a lot of it right now though.

PR #9, opened 2020-09-16
Allow specification of a client name or set of scopes on Blueprint

jimPruyne:
THis allows the user creating an ActionProviderBlueprint to customize
somewhat the authentication process by specifying a (resource server)
named to be used when validating input Globus Auth tokens, and to
specify a set of scopes which are also considered valid for all
operations via a Blueprint.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ venv.bak/
 
 # whattimeisitrightnow example provider's database file
 whattimeisitrightnow.example.db
+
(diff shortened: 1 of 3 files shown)


urimandujano (approved):
Good stuff here - I especially like the keyword-arg only addition to the ActionProviderBlueprint.  It's a good way of making this backwards-compatible. I also appreciate the general cleanup pass!

PR #10, opened 2020-09-16
Prepare for 0.10.1 release

jimPruyne:


urimandujano (changes requested):
Can we also update the version in globus_action_provider_tools/__init__.py? I don't think it's used anywhere but since it's there it would be good to keep it up to date.

jimPruyne:
> Can we also update the version in globus_action_provider_tools/**init**.py? I don't think it's used anywhere but since it's there it would be good to keep it up to date.

Done. I'll have to get used to making the update there as well. Thanks for letting me know.

urimandujano (approved):

PR #11, opened 2020-09-21
Removes shared methdispatch implementation

urimandujano:
There is a subtle bug in the existing implementation where if multiple (>=2) `ActionProviderBlueprints` are created and attached to a Flask app, the implementation of `methdispatch` would only ever record one Blueprint's `status`, `cancel` and `release` endpoints. For example, this meant that all requests to a `status` operation on the app would get routed to one of the `ActionProviderBlueprint`'s routes (typically the last Blueprint to get registered on the app). What should happen is that calls to a Blueprint's endpoint should get routed to that Blueprint's routes.

This PR fixes the issue by removing the `methdispatch` implementation. Aside from the buggy implementation, it was hard to understand and hard to maintain. It's purpose was to transparently call the correct route depending on if the route was called with a `str` or `ActionStatus` argument. Without it, it's a bit more digestable to decribe the behavior as "if an action_loader is registered, your endpoint routes should expect an ActionStatus type value as the first argument".

Also pdates type annotations to keep mypy happy while also providing useful feedback to devs.

### DIFF
--- a/globus_action_provider_tools/flask/apt_blueprint.py
+++ b/globus_action_provider_tools/flask/apt_blueprint.py
@@ -1,9 +1,6 @@
 import functools
-import inspect
 import json
-import logging
-from functools import singledispatch, update_wrapper
-from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union, overload
 
 from flask import Blueprint, Response, current_app, g, jsonify, request
 from jsonschema.validators import Draft7Validator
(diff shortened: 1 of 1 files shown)


globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -217,22 +177,31 @@ def wrapper() -> ViewReturn:
...
         """
         Decorates a function to be run as an Action Provider's status endpoint.
         """
-        _, first_param = list(inspect.signature(func).parameters.items())[0]
-        if first_param.annotation == first_param.empty:
-            # There's no annotation available, assume the function handles str
-            self._generic_action_status.register(str, func)
-        else:
-            self._generic_action_status.register(first_param.annotation, func)
+        self._action_status: ActionStatusType = func
```
jimPruyne, line 204: Any preference for initializing these sorts of values to `None` in ``__init__`` and then doing an `is None` check rather than `except AttributeError`? I guess the attribute error is considered pythonic under forgiveness/not permission, but I float the alternative. I'm fine either way.

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -217,22 +177,31 @@ def wrapper() -> ViewReturn:
         print(f'Registered action run plugin "{func.__name__}"')
         return wrapper
 
-    @methdispatch
-    def _generic_action_status(self, *args, **kwargs):
-        raise NotImplemented(
-            f"No action_status function registered to handle type {type(args[0])}"
-        )
+    @overload
```
jimPruyne, line 180: Another new python feature for me. This is kinda a weird syntax, not that there's anything that can be done about it.

How does mypy handle things if the first argument is just listed as a `Union` type rather than adding the `overload` multi-definition kinda thing?

jimPruyne (commented):
So, this does look like a cleaner implementation in the end. After reading the description, then reading the code, and now reading the description again, my comment in-line related to use of `Union` might not make sense, but I'll leave it in place in case I'm still missing something subtle here.

I guess I'm not sure why the original implementation has this bug (i'd have to read the now removed stuff more carefully I guess, and I tended to overlook the removed parts), but given that it is gone and as both you and I said, this implementation looks pretty simple, I'm ok not going into that detail.

So, my comments are pretty minor, mostly for my own clarification and to convince myself that my idea around `Union` is bad.

Good job catching a subtle bug!

globus_action_provider_tools/flask/apt_blueprint.py
urimandujano: That's one of the first things I tried (I really didn't like the idea of creating function stubs just for using `@overload`) but mypy apparently tries to do an AND between the types declared in the `Union` so that mypy reports errors for both `str` and `ActionStatus` typed args. The only alternative was using typing Protocols but the mypy errors they generate are opaque. I figured actionable mypy errors > neatly defined typing definitions.

urimandujano:
I'm not entirely sure myself how to bug in the faulty implementation worked. I think it has to do with how python marries decorators to class methods - it ended up being that all instances shared the same `methdispatch` reference. Removing it is one less thing to keep track of.

PR #14, opened 2020-10-01
Updates mypy annotation to support Iterables in monitor_by and manage_by

urimandujano:
This PR updates mypy annotations to accept any `Iterable` value for `manage_by` and `monitor_by`, effectively allowing Lists, sets, and frozensets, to be provided as values without errors. It also updates the `ActionProviderJSONEncoder` such that it can serialize types of `AbstractSet`. 

Examples are also updated.


jimPruyne (approved):
I think this looks good. If after reading this and thinking for a minute I've figured that you're replacing the `+` with `chain` so you can concatenate arbitrary iterables (e.g. list and sets) that makes sense in the context of this changeset.

So, if you see the review and confirm that's the thinking, then go ahead and consider this done and merge is ok by me.

urimandujano:
Yup you nailed it - using `chain` instead of `+` allows me to combine arbitrary iterables.

PR #15, opened 2020-10-05
Updates and documents testing tools

urimandujano:
Improves the state of testing tools in ActionProviderTools. The `mocks.py` module is greatly slimmed down and provides a new `TokenChecker` mock which can be used to replace `TokenChecker` instances.  The addition of `fixtures.py` introduces fixtures that mock Globus Auth integration for the two Flask toolk, APTBlueprint and Flask Helpers.

Examples are updated to demonstrate using the fixtures to test apps without a Globus Auth Client ID, secret, or token.

Docs are updated with to reflect the new fixtures and updated mocks as well as with information on when/how to use each fixture and when to use the mocks instead.

Adds a deprecation warning to the testing patches, mostly because the `flask_blueprint_tokenchecker_patch` is not as good as the `apt_blueprint_noauth` fixture (the fixture mocks only the blueprint-under-test's TokenChecker whereas the patch mocks the TokenChecker globally for all blueprint's.). Also the `flask_api_helpers_tokenchecker_patch` is effectively the same as the `flask_helpers_noauth` fixture, except in both cases the fixtures are more convenient and easier to work with.

Adds tests validating the new mocks. "Dogfoods" the new fixtures by using them in the toolkit test's for the Flask helpers.

### DIFF
--- a/docs/source/toolkit/testing.rst
+++ b/docs/source/toolkit/testing.rst
@@ -1,0 +1,213 @@
+Testing
+=======
+
+An Action Provider is closely tightle with Globus Auth (see
+:ref:`globus_auth_setup`). This integration makes it easy to validate incoming
+requests and ensures that the requestor is authorized to execute actions against
+an Action Provider. However, the integration can make it difficult to run tests
+against an Action Provider to validate that its endpoints behave correctly.
+During a CI/CD pipeline, it may be desirable to start your Action Provider
+without a valid Client ID or Secret just to make sure that the app behaves
+correctly.
+
+The toolkit provides various tools to enable testing and validation in the
+:code:`globus_action_provider_tools.testing` module.
+
+
+Fixtures
+^^^^^^^^
+
+The toolkit provides various `pytest fixtures
+<https://docs.pytest.org/en/stable/fixture.html>`_  that greatly reduce the need
+to manually mock and patch interactions with Globus Auth. If an Action Provider
+is created using any of the Flask helpers provided in the
+:code:`globus_action_provider_tools.flask` module, we provide a fixture to
+easily mock authentication out of your Action Provider. Each of the Flask
+helpers internally creates a *TokenChecker* which does two things: it validates
+that the Action Provider is correctly configured with a valid Globus Auth Client
+ID and Secret, and it validates that incoming requests contain a valid token.
+These fixtures abstract both aspects of the internal *TokenChecker* so that you
+can focus on testing your Action Provider's behavior and logic. 
+
+
+Action Provider Blueprint
+-------------------------
+
+If your Action Provider is built using the *ActionProviderBlueprint* Flask
+helper, use the :code:`apt_blueprint_noauth` fixture in
+:code:`globus_action_provider_tools.testing.fixtures`:
+
+.. code-block:: python
+                
+    from globus_action_provider_tools.testing.fixtures import (
+        apt_blueprint_noauth
+    )
+
+Once imported, it can be used just as any other pytest fixture. We recommend
+passing it as a parameter to another fixture which ultimately creates the app
+and returns a resource for use in tests. Provide the fixture your
+*ActionProviderBlueprint* instance under test. This will update your instance
+with stubbed out authentication. The example below shows how to create a
+:code:`client` fixture which can be used to make unauthenticated HTTP requests
+against the Action Provider:
+
+.. code-block:: python
+
+    from myapp import action_provider_blueprint
+
+    @pytest.fixture(scope="module")
+    def client(apt_blueprint_noauth):
+        apt_blueprint_noauth(action_provider_blueprint)
+        app = create_app()
+        return app.test_client()
+
+Once composed like this, you can use the :code:`client` fixture in your tests to
+receive and use a Flask *test_client* to make unauthenticated requests against
+your Action Provider:
+
+.. code-block:: python
+
+    def test_introspection_endpoint(client):
+        response = client.get("/")
+        assert response.status_code == 200
+
+
+Flask API Helpers
+-----------------
+
+If your Action Provider is built using the
+:code:`add_action_routes_to_blueprint` Flask helper, use the
+:code:`flask_helpers_noauth` fixture in
+:code:`globus_action_provider_tools.testing.fixtures*`:
+
+.. code-block:: python
+                
+    from globus_action_provider_tools.testing.fixtures import (
+        flask_helpers_noauth
+    )
+
+Once imported, simply pass the :code:`flask_helpers_noauth` fixture as a
+parameter to another fixture which creates the app and returns a resource for
+use in tests. Unlike the :code:`apt_blueprint_noauth` fixture, the
+:code:`flask_helpers_noauth` fixture does not need to be explicitly executed -
+simply passing it as a parameter is sufficient to temporarily stub out the
+Provider's authentication. An example of how to create a :code:`client` fixture
+which can be used to make unauthenticated HTTP requests against the Action
+Provider is shown below:
+
+.. code-block:: python
+
+    @pytest.fixture(scope="module")
+    def client(flask_helpers_noauth):
+        app = create_app()
+        app.config["TESTING"] = True
+        yield app.test_client()
+
+Once composed like this, you can use the :code:`client` fixture in your tests to
+receive and use a Flask *test_client* to make requests against your Action
+Provider:
+
+.. code-block:: python
+
+    def test_introspection_endpoint(client):
+        response = client.get("/")
+        assert response.status_code == 200
+
+.. note:: 
+    
+    The :code:`flask_helpers_noauth` fixture will patch the TokenChecker in a
+    global scope during testing, meaning that any other Action Providers that
+    are themselves built using the Flask API Helpers will also have their
+    TokenChecker's patched. This may lead to unintended issues if testing
+    multiple Action Providers in the same pytest test session. If this is your
+    case, we highly recommend isolating your Action Provider tests.
+
+
+Mocks
+^^^^^
+
+The toolkit provides various `mocks
+<https://docs.python.org/3/library/unittest.mock.html#the-mock-class>`_ which
+can be used individually to stub out your Action Provider's authentication. You
+should use these directly if you are writing an Action Provider using a
+non-Flask framework or if you've decided not to use the built in Flask helpers.
+
+.. note::
+
+    This toolkit uses these mocks within the
+    :code:`globus_action_provider_tools.testing.fixtures` module.  
+
+
+.. _mock-authstate:
+
+Mock AuthState
+--------------
+
+An *AuthState* represents a requestor's authentication status and Globus Auth
+information. Every request should have its token validated via the
+*TokenChecker*'s :code:`check_token` method, which in turns generates an
+*AuthState* object.
+
+During testing, it is desirable to not need to provide valid tokens with every
+request. Use the :code:`mock_authstate` mock to generate a stubbed out
+*AuthState* object that won't validate requestor properties against Globus Auth.
+This is most useful when used in a patch as the return value for the
+*TokenChecker*'s :code:`check_token` method:
+
+.. code-block:: python
+
+    from unittest import mock     
+
+    import pytest
+    from globus_action_provider_tools.testing.mocks import mock_authstate
+
+    ...
+
+    @pytest.fixture
+    def client():
+        with mock.patch(
+            "globus_action_provider_tools.authentication.TokenChecker.check_token",
+            return_value=mock_authstate(),
+        ):
+            yield app.test_client()
+
+The example above creates a fixture which can be used to create a client that
+can make unauthenticated HTTP requests against an Action Provider. 
+
+
+Mock TokenChecker
+-----------------
+
+Because the *TokenChecker* is this toolkit's authentication workhorse, it's
+possible to entirely replace the the *TokenChecker* with a mock object. Doing so
+will allow your Action Provider to start up without validating its Client ID or
+Secret and will also allow unauthenticated requests to be made against it. This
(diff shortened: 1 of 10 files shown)


docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
...
+
+.. code-block:: python
+
+    def test_introspection_endpoint(client):
+        response = client.get("/")
+        assert response.status_code == 200
+
+.. note:: 
+    
+    The :code:`flask_helpers_noauth` fixture will patch the TokenChecker in a
```
jimPruyne, line 118: Seems less likely for others than how we do things, but an excellent note to make it clear.

docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Mock AuthState
+--------------
+
+An *AuthState* represents a requestor's authentication status and Globus Auth
+information. Every request should have its token validated via the
+*TokenChecker*'s :code:`check_token` method, which in turns generates an
+*AuthState* object.
+
+During testing, it is desirable to not need to provide valid tokens with every
```
jimPruyne, line 151: "no not need to" perhaps awkward? (Am I sounding like an English teacher...)

docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
...
+request. Use the :code:`mock_authstate` mock to generate a stubbed out
+*AuthState* object that won't validate requestor properties against Globus Auth.
+This is most useful when used in a patch as the return value for the
+*TokenChecker*'s :code:`check_token` method:
+
+.. code-block:: python
+
+    from unittest import mock
```
jimPruyne, line 159: If we're using pytest, does using its monkeypatch module make at least as much sense as using `unittest.mock`? Do we have concerns about mixing unittest and pytest?

docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
...
+
+    from unittest import mock     
+
+    import pytest
+    from globus_action_provider_tools.testing.mocks import mock_authstate
+
+    ...
+
+    @pytest.fixture
+    def client():
```
jimPruyne, line 167: With monkeypatch, I think this would look roughly like:
```python
@pytest.fixture
def client(monkeypatch):
    monkeypatch.settattr("globus_action_provider_tools.authentication.TokenChecker.check_token",
            mock_authstate)
```

So pretty similar.

docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
...
+            "globus_action_provider_tools.authentication.TokenChecker.check_token",
+            return_value=mock_authstate(),
+        ):
+            yield app.test_client()
+
+The example above creates a fixture which can be used to create a client that
+can make unauthenticated HTTP requests against an Action Provider. 
+
+
+Mock TokenChecker
```
jimPruyne, line 178: This seems like a really good approach for non-flask apps? Should we push this description forward so people get to it before the other approach?

docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
...
+            app.config["TESTING"] = True
+            yield app.test_client()
+
+
+.. note::
+
+    This example will only work if there's a function or method that is used to
+    create the TokenChecker instance. It demonstrates how you can patch a
+    function or a method to return the Mock TokenChecker. Internally, the Mock
+    TokenChecker will generate the :ref:`mock-authstate` objects described
```
jimPruyne, line 212: Ah, this kinda makes sense to have the other description first. Perhaps makes my previous comment not so good.

examples/apt_blueprint/tests/test_action_provider.py
```diff
@@ -3,21 +3,19 @@
...
-    with flask_blueprint_tokenchecker_patch:
-        app = create_app()
-        yield app.test_client()
+def client(apt_blueprint_noauth):
+    # Patch the blueprint BEFORE attaching it to an app so that the Provider can
+    # start without valid credentials AND requests can be made without supplying
+    # a valid token
+    apt_blueprint_noauth(aptb)
+    app = create_app()
+    return app.test_client()
```
jimPruyne, line 18: The examples in the docs use `yield` here its `return`. GIven that there's no code after the `yield` above, there's probably no actual difference (?). Maybe nice to be consistent. I think we get used to just using `yield` in fixtures sometimes.

examples/watchasay/tests/test_action_provider.py
```diff
@@ -21,20 +21,17 @@
...
-        app = create_app()
-        app.config["TESTING"] = True
-        yield app.test_client()
+def client(flask_helpers_noauth):
+    # Patch the app using the flask_helpers_noauth fixture so that the Provider
+    # can start up without valid credentials AND requests can be made without
+    # supplying a valid token
+    app = create_app()
+    app.config["TESTING"] = True
+    yield app.test_client()
```
jimPruyne, line 34: See above for yield vs return

examples/whattimeisitrightnow/tests/test_action_provider.py
```diff
@@ -31,14 +31,20 @@
...
 
 @pytest.fixture
 def client():
-    app.config["TESTING"] = True
-    yield app.test_client()
+    with mock.patch(
+        "globus_action_provider_tools.authentication.TokenChecker.check_token",
+        return_value=mock_authstate(),
+    ):
+        from examples.whattimeisitrightnow.app.app import app
```
jimPruyne, line 44: is this import here related to how the mock is getting installed? Its kinda a funky notation...

globus_action_provider_tools/testing/fixtures.py
```diff
@@ -0,0 +1,42 @@
...
+    Auth integration. The fixture returns a function to which the instance
+    should be supplied as a parameter:
+
+    i.e. apt_blueprint_noauth(aptb)
+    """
+
+    def _apt_blueprint_noauth(aptb: ActionProviderBlueprint):
+        # Manually remove the function that creates the internal token_checker
+        for f in aptb.deferred_functions:
+            if f.__name__ == "_create_token_checker":
```
jimPruyne, line 22: Given how this is mucking with the internals, really good to provide the function for people as we wouldn't want them to have to figure this out.

globus_action_provider_tools/testing/fixtures.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pytest.fixture(scope="module")
+def flask_helpers_noauth():
+    """
+    A fixture designed to mock the Globus Auth integration in an Flask app
+    created using the api_helpers. Simply using this fixture will allow creating
+    the mocked app.
+    """
+    with patch(
```
jimPruyne, line 38: So, I see that my suggestion around monkeypatch might have repercussions down in the internal implementation. Maybe take a quick look to see if it does it in a nice way that keeps us consistent in the pytest world since we're already assuming it. If it makes it any bit harder, don't worry about it.

globus_action_provider_tools/testing/fixtures.py
```diff
@@ -0,0 +1,42 @@
...
+        "globus_action_provider_tools.flask.api_helpers.TokenChecker.check_token",
+        return_value=mock_authstate(),
+    ):
+        yield
```
jimPruyne, line 42: one nice thing about monkeypatch is you don't have to do this yield. Since the monkeypatch itself is a fixture, it handles that.

globus_action_provider_tools/testing/mocks.py
```diff
@@ -1,45 +1,41 @@
-from typing import NamedTuple
-from unittest.mock import PropertyMock, patch
+from unittest.mock import Mock
```
jimPruyne, line 1: So, here, unittest.mock makes sense to me as it doesn't really on any outside packages to be installed so people not going flask-y can also not go pytest-y.

Either I'm over thinking this, or under thinking it. I'm not quite sure...

globus_action_provider_tools/testing/mocks.py
```diff
@@ -1,45 +1,41 @@
...
+    auth_state.check_authorization.return_value = True
+    auth_state.introspect_token.return_value = None
 
-    auth_client = ConfidentialAppAuthClient(None, None)
-    auth_state = AuthState(auth_client, None, None)
-    auth_state.check_authorization = mock_check_authorization
-    auth_state.primary_identity = mock_primary_identity
-    auth_state.introspect_token = mock_introspect_token
-    auth_state.bearer_token = "MOCK_BEARER_TOKEN"
+    return auth_state
```
jimPruyne, line 31: This looks much cleaner, nice!

globus_action_provider_tools/testing/patches.py
```diff
@@ -1,7 +1,19 @@
+import warnings
 from unittest.mock import patch
 
 from globus_action_provider_tools.testing.mocks import mock_authstate
 
+warnings.warn(
```
jimPruyne, line 6: nice

tests/test_flask_helpers/conftest.py
```diff
@@ -28,7 +32,7 @@
 
 
 @pytest.fixture()
-def aptb_app(auth_state):
+def aptb_app(apt_blueprint_noauth, auth_state):
```
jimPruyne, line 35: fixture here is so nice to read compared to the `with patch`

jimPruyne (commented):
Looks really good. Really nice to see both the functionality and particularly the full docs!

A bit of general wording stuff in docs, but pretty minor.

Some quesitons about using unittest.mock vs pytest.monkeypatch in places? But, I'm open to whichever solution between those two seems best to you.

docs/source/toolkit/testing.rst
```diff
@@ -0,0 +1,213 @@
+Testing
+=======
+
+An Action Provider is closely tightle with Globus Auth (see
```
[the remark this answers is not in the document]
urimandujano: Wow, I can't remember what I was going for. Switching to "integrated" since it sounds like a fancier word.

docs/source/toolkit/testing.rst
urimandujano: Making note of things that have bitten me before so that other's don't get hurt (:

docs/source/toolkit/testing.rst
urimandujano: I think the only concern is providing a uniform non-confusing story for folks who are ramping up with testing/using the tools. `unittest` and `pytest` play pretty well with each other so I'm not too worried about them interfering with each other (unless there are cases and I just haven't run across them yet). 

There are some gotchas with using pytest's `monkeypatch` fixture because it's function-scoped. The fixtures provided here are using `patch` and `mock` directly so they can be used in fixtures regardless of their scope. I like the idea of using `monkeypatch` in the examples since it's syntax is clearer than patch's.

docs/source/toolkit/testing.rst
urimandujano: Thanks for the working example!

docs/source/toolkit/testing.rst
urimandujano: I haven't had the chance to test this on a non-Flask app to see how we could just mock the `TokenChecker` import so I'm not sure how/if it'll work. I think mocking the import would work but I'm not quite ready to claim that it's possible or that it's a good idea.

For now I've added it as a piece that can be used and outlined a particular way of using it that I know works (via the provided fixtures) we start experimenting with other frameworks or want to revisit how framework agnostic our tools are, this would be something to work on pushing forward.

examples/apt_blueprint/tests/test_action_provider.py
urimandujano: I've also seen yield most often used in test fixtures so `yield` it is!

examples/whattimeisitrightnow/tests/test_action_provider.py
urimandujano: Yea it has to do with how the `TokenChecker` gets instantiated at import time... The example could probably be refactored to have a slightly cleaner way of getting mocked but I was hoping not to spend too much time refactoring examples.

globus_action_provider_tools/testing/fixtures.py
urimandujano: Yea I think using `monkeypatch` internally would impose some limitations on how our libraries fixtures could be used (i.e. they could only be using in a certain scope). Looking at this again I've actually decided to broaden the scope of our fixtures to "session" (which is as I understand it the broadest scope possible) so that they can be used in fixtures of any scope. 

Using patches at this level maintains allows some extra flexibility for use cases, which may or may not be needed at this point.

globus_action_provider_tools/testing/mocks.py
urimandujano: I'm not sure how to prove that it's over or under thinking (: My hope in providing base mocks was that they'd be able to be composed by higher level tools depending on what needed to be accomplished. I think it worked nicely that the `fixtures` end up dogfooding the `mocks` and that the `tokenchecker` mock dogfoods the `auth_state` mock. From some of my udpates to the examples, they also seem to fit in well when using `monkeypatch`.

urimandujano:
Thank you for the review! It was good for me to step back and think through the strengths of `unittest's` tools and `monkeypatch`. I think I've convinced myself that at the lowest level, unittest tools provide the most flexibility (which is welcome since we're trying to provide general purpose testing tools) but if that flexibility isn't necessary, `monkeypatch` is the way to go (useful for the testing examples + realworld usage).

Really interesting to me is that it seems like there are lots of people trying to make sense of the two mock methods and opinions vary widely: https://github.com/pytest-dev/pytest/issues/4576.

jimPruyne (approved):
After discussion, this all looks good.

PR #17, opened 2020-10-09
Adds GH Action to release package

urimandujano:
Everytime we create (or edit) a release, this Action will automatically build and publish the release to PyPi.

Note that the build will publish the release according to what is in pyproject.toml. So we would probably still want to go through the formalities of creating and merging PR branches with updates preparing for a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,29 @@
+name: PyPi Publish
+
+on:
+  release:
+    types: [created, edited]
+
+jobs:
+  publish:
+    name: Publish a release to PyPi
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


jimPruyne (approved):
I think I like the idea on this. I've still done very little with GH Actions, so I cannot fully check the syntax here. But, it look reasonable.

What credentials are you using to publish to pypi? I see the values there, but I haven't gone to see what secrets there are etc. And, I'm not sure what the `__token__` syntax does.

So, I'm approving for now, but we'll probably want to run through this once together to see how it really works.

urimandujano:
I generated a PyPi token for this repo with upload permissions and that's the token that's getting pulled into the release pipeline. And the `username: __token__`  syntax is just how you indicate to PyPi that you're using a token as a credential for uploading a package -- https://pypi.org/help/#apitoken.

PR #20, opened 2020-12-17
Implements standardized ActionProvider exceptions

urimandujano:
Adds werkzeug-based HTTP exceptions that can be raised to return JSON error results containing http status code, error name, and error description. The error description can be customized by the user.

Subclasses the built in `str` class in our enum classes to allow string comparisons to our enum classes to succeed without the need to access the `.value` attribute.

Also refactors the Flask helpers to improve shared methods and types between the two Flask helper tools.

### DIFF
--- a/globus_action_provider_tools/exceptions.py
+++ b/globus_action_provider_tools/exceptions.py
@@ -1,0 +1,76 @@
+"""
+exceptions.py contains JSON-able HTTP exceptions based off the werkzeug
+HTTPException class. These custom exceptions produce a JSON response containing
+error name, http status code, and a description of the issue. 
+
+These exceptions can be raised with a default message directly from werkzeug:
+    raise ActionProviderToolsException
+
+or a custom error message can be supplied:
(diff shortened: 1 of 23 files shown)


globus_action_provider_tools/exceptions.py
```diff
@@ -0,0 +1,76 @@
+"""
```
jimPruyne, line 1: Just dropping this comment at the top. This looks like a pretty nice method of unifying errors based on exceptions raised.

globus_action_provider_tools/flask/types.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, Callable, Dict, Sequence, Set, Tuple, Union
```
jimPruyne, line 1: Good seeing this unified in one place.

jimPruyne (approved):
Looks like a lot of changes, but it looks good to me. Let's go with it as is!

PR #21, opened 2021-01-04
Ignores trailing slashes on introspection

urimandujano:
Modifies the `ActionProviderBlueprint` so that the introspection endpoint is available at the root path, regardless of if the request contains a trailing "/". This only affects the `ActionProviderBlueprint` since the Flask action_routes helper already implements this.
[ch5659]

### DIFF
--- a/globus_action_provider_tools/flask/apt_blueprint.py
+++ b/globus_action_provider_tools/flask/apt_blueprint.py
@@ -105,7 +105,11 @@ class ActionProviderBlueprint(Blueprint):
         self.record_once(self._create_token_checker)
 
         self.add_url_rule(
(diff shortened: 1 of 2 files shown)


urimandujano:
@jimPruyne - I opted to only implement this on the introspection endpoint since it seemed like the biggest source of confusion/inconsistency. I thought that if a user really needs to behavior on other endpoints, they can set application wide by setting `app.url_map.strict_slashes = False` which is both more flexible and less code for us.

jimPruyne (commented):
Can you confirm that this still works on GET without the trailing slash? If so, this looks fine for now and I agree that this is the most common source of problems that I've seen.

urimandujano:
Yup I can confirm that!

PR #22, opened 2021-01-06
Adds support for pydantic based input validation

urimandujano:
Extends Flask tools' capability to use pydantic models when creating and validating request input, `ActionRequests`, and `ActionStatuses`. Developers can choose to use pydantic models or write jsonschema definitions by hand.

The general idea is that when an `ActionProvider` starts up, it inspects its `ActionProviderDescription` to determine if its input schema is a `str`, `dict`, or `pydantic` model. Depending on the type of its input schema, the appropriate validation function (json-schema or pydantic based) is loaded and used to validate incoming requests. Since the `ActionRequest` and `ActionStatus` objects are now pydantic models which throw `ValidationError`s if created with the wrong data, the Flask helpers now run the user supplied callbacks in try-except blocks and log errors/raise exceptions appropriately.

Refactors a bit to stay DRY and adds tests + docs.

[ch5660]


examples/whattimeisitrightnow/app/app.py
```diff
@@ -37,23 +37,8 @@
...
-            iterable = iter(obj)
-        except TypeError:
-            pass
-        else:
-            return list(iterable)
-        return super().default(obj)
-
-
-app.json_encoder = CustomJSONEncoder
+
```
jimPruyne, line 40: This is nice, removing all that encoder logic.

globus_action_provider_tools/data_types.py
```diff
@@ -84,31 +86,93 @@ class ActionProviderDescription:
...
-    action_id: str = field(default_factory=shortish_id)
-    start_time: str = field(default_factory=now_isoformat)
-    label: Optional[str] = None
-    monitor_by: Optional[Iterable[str]] = None
-    manage_by: Optional[Iterable[str]] = None
-    completion_time: Optional[str] = None
-    release_after: Optional[str] = None
-    display_status: Optional[str] = None
-    details: Optional[Dict[str, Any]] = None
+class ActionRequest(BaseModel):
```
jimPruyne, line 89: If this is no longer a dataclass, do you know if our code in our actual providers which I believe uses the dataclass-based method of creating a dict for storage in dynamo will still work?

pydantic provides a dataclass implementation as well which I think blends fine with their BaseModel. Perhaps decorating with the pydantic dataclass implementation gets us both behaviors we want.

globus_action_provider_tools/data_types.py
```diff
@@ -84,31 +86,93 @@ class ActionProviderDescription:
...
+    request_id: str = Field(
+        ...,
+        description="A unique identifier representing the request to start an action. Multiple uses of the same request_id must have the same content or they will be rejected. Only one instance of the operation will be executed, so requests with the same request_id may be repeated to attempt to guarantee execution of an action",
```
jimPruyne, line 92: Nice that you can add description fields to all these!

globus_action_provider_tools/data_types.py
```diff
@@ -84,31 +86,93 @@ class ActionProviderDescription:
...
+        max_length=64,
+    )
+    deadline: Optional[datetime.datetime] = Field(
+        None,
+        description="A timestamp indicating by which time the action must complete. The request may be rejected if the Action Provider does not expect to be able to complete the action before the deadline or if it represents a time greater than the maximum_deadline specified in the Provider Description.",
+    )
+    release_after: Optional[str] = Field(
+        None,
+        description="An ISO8601 time duration value indicating how long retention of the status of the action be retained after it reaches a completed state. Action Providers may limit the maximum value. It is recommended that Providers provide a default release_after value of approximately 30 days.",
+        regex="^P(?!$)(\d+(?:\.\d+)?Y)?(\d+(?:\.\d+)?M)?(\d+(?:\.\d+)?W)?(\d+(?:\.\d+)?D)?(T(?=\d)(\d+(?:\.\d+)?H)?(\d+(?:\.\d+)?M)?(\d+(?:\.\d+)?S)?)?$",
```
jimPruyne, line 111: And nice that the regex can be added here

globus_action_provider_tools/data_types.py
```diff
@@ -84,31 +86,93 @@ class ActionProviderDescription:
...
+    )
+    start_time: str = Field(default_factory=now_isoformat)
+    label: Optional[str] = Field(
+        None,
+        description="A short human presentable description of the Action requested",
+        min_length=1,
+        max_length=64,
+    )
+    monitor_by: Set[str] = Field(
+        default_factory=list,
```
jimPruyne, line 148: If this is type `Set` should the `default_factory` also be `set`?

globus_action_provider_tools/data_types.py
```diff
@@ -84,31 +86,93 @@ class ActionProviderDescription:
...
+        description="A list of principal URNs containing identities which are allowed to monitor the progress of the action using the /status and /log operations. When not provided, defaults to the user that initiated the action.",
+        regex="^urn:globus:(auth:identity|groups:id):([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})$",
+    )
+    manage_by: Set[str] = Field(
+        default_factory=list,
+        description="A list of principal URNs containing identities which are allowed to manage the progress of the action using the cancel and release operations. When not provided, defaults to the user that initiated the action.",
+        regex="^urn:globus:(auth:identity|groups:id):([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})$",
+    )
+    completion_time: Optional[datetime.datetime] = Field(
```
jimPruyne, line 157: Does using type `datetime.datetime` get us any iso representation validation via pydantic? Should there be a regex of some sort if not (though perhaps that's not good since it is not a `str` type)

globus_action_provider_tools/data_types.py
```diff
@@ -118,8 +182,12 @@ class ActionProviderJsonEncoder(JSONEncoder):
     def default(self, obj):
         if is_dataclass(obj):
             return asdict(obj)
-        elif isinstance(obj, Enum):
-            return obj.name
         elif isinstance(obj, AbstractSet):
             return list(obj)
+        elif isinstance(obj, BaseModel):
+            return obj.dict()
+        elif inspect.isclass(obj) and issubclass(obj, BaseModel):
```
jimPruyne, line 189: Ah, so here's the actual thing that makes this happen. Looks pretty nice.

globus_action_provider_tools/flask/api_helpers.py
```diff
@@ -241,44 +212,64 @@ def action_run() -> ViewReturn:
...
-
-        if result.error_msg is not None:
-            raise BadActionRequest(result.error_msg)
+        action_request = validate_input(
+            request.get_json(force=True), input_body_validator
+        )
 
-        request_json = request.get_json(force=True)
+        # It's possible the user will attempt to make a malformed ActionStatus -
+        # pydantic won't like that. So log and handle the error with a 500
```
jimPruyne, line 220: This seems like a reasonable course of action in this case.

globus_action_provider_tools/flask/helpers.py
```diff
@@ -105,3 +117,80 @@ def blueprint_error_handler(exc: Exception) -> ViewReturn:
...
+
+
+def validate_input(
+    request_json: Dict[str, Any], input_body_validator: ActionInputValidatorType
+) -> ActionRequest:
+    """
+    Ensures the incoming request conforms to the ActionRequest schema and
+    the user-defined ActionProvider input schema.
+    """
+    try:
```
jimPruyne, line 129: This looks nice and clean for validation on request.I like it.

globus_action_provider_tools/flask/helpers.py
```diff
@@ -105,3 +117,80 @@ def blueprint_error_handler(exc: Exception) -> ViewReturn:
...
+        action_request = ActionRequest(**request_json)
+    except ValidationError as ve:
+        raise BadActionRequest(ve.errors())
+
+    input_body_validator(action_request.body)
+    return action_request
+
+
+def get_input_body_validator(
```
jimPruyne, line 138: Just placing this at the top of the function: this looks nice and clean as well. Another key part to how this holds together and it looks really good.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ pyyaml = "^5.3.1"
 arrow = "^0.15.5"
 pybase62 = "^0.4.0"
 dataclasses = {version = "^0.7", python = "~3.6"}
+pydantic = "^1.7.3"
```
jimPruyne, line 40: Another thing if we switch to the pydantic dataclass implementation, we can probably get rid of the dataclass import from the previous line of t his.

jimPruyne (commented):
This all looks really, really nice. Main idea, if I recall correctly, is to use pydantic dataclass implementation which should help across python versions (I think) and keep dataclass behavior we may rely on in some of our other tools.

Given how much change is here, including docs and tests, this is quite excellent.

docs/source/examples/input_schemas.rst
```diff
@@ -0,0 +1,119 @@
+Action Provider Input Schemas
+=============================
+
+We support user-written JSONSchema and Pydantic_ based schema defintions. In
```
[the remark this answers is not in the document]
urimandujano: In this one I tried out a different way of making rst links - I added `.. _Pydantic: https://pydantic-docs.helpmanual.io/` to the bottom of the doc so that hopefully the link appears once it renders.

docs/source/examples/input_schemas.rst
```diff
@@ -0,0 +1,119 @@
...
+
+JSONSchema
+^^^^^^^^^^
+
+A typical schema definition may look like:
+
+.. code-block:: JSON
+
+    {
+        "$id": "https://automate.globus.org/skeleton_action_provider.input.schema.json",
```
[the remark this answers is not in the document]
urimandujano: Ah so that's what the field gives us. I think there's a way to do that on the pydantic BaseModels. I'll need to think what the best way to add that to user input models will be. Do you think this applies more to one JSONSchema than another? I can think of the jsonschema presented during introspection, the action request schema, and the action provider input schema.

docs/source/examples/input_schemas.rst
```diff
@@ -0,0 +1,119 @@
...
+            ...,
+            title="Echo String",
+            description="An input value to this ActionProvider to echo back in its response",
+        )
+
+        # pydantic lets you display examples of passing input 
+        class Config:
+            schema_extra = {"example": {"echo_string": "hi there"}}
+
+With the model created, pass the **class name** to the ActionProvider
```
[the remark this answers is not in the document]
urimandujano: Good clarification, this is a little weird so I want to make sure the instructions are as clear as possible.

globus_action_provider_tools/data_types.py
urimandujano: This is a good question, I just did a search on `globus-automate` to remind myself how we're dumping the `ActionStatus` objects and there's two things going on -> we `asdict()` it and dump it using the `ActionProviderJsonEncoder`. If we're using the custom encoder, we should be good on that since it knows how to handle pydantic types. The `asdict` part I'm not so sure about but we can get rid of that since the custom encoder also handles dumping dataclasses.

globus_action_provider_tools/data_types.py
urimandujano: I was impressed that the regex field is actually used to validate incoming strings. So apart from being documentation, it does meaningful validation!

globus_action_provider_tools/data_types.py
urimandujano: I'm on the fence about this one, pydantic interprets the set type annotation to mean it should be a sequence of unique items, and the validation will make sure of that. The default factory is ends up being the code representation and since theyre being used in different ways, I'm not sure about the best way to go forward. Let me see if setting this to be a set factory breaks anything.

globus_action_provider_tools/data_types.py
urimandujano: Yup, datetimes will be parsed as iso8601. I think pydantic supports a few different formats when converting to a datetime, and iso8601 is just one of them. I hope that the description of the field is enough to put a user on the right path.

pyproject.toml
urimandujano: Our ActionProviderDescription is the last thing that uses the dataclass decorator. For portability it may make sense to remove the use there and get rid of the package altogether? If we switch it over to the pydantic model type we get some field validation (which may be useful since we're expecting to interpret this on introspection) but otherwise dataclasses and pydantic models behave pretty similarly. I'm open to either direction.

im reminded of when id get tripped up updating things bc of the dataclass dependency, and that faint memory makes me want to remove it.

globus_action_provider_tools/data_types.py
urimandujano: I just tried this out, if we try to run asdict() on an instance of a `pydantic` `ActionStatus`, we get `TypeError: asdict() should be called on dataclass instances`. We're good if we just remove the call to `asdict` in the `json.dump`.

urimandujano:
@jimPruyne gave this an update based on the review we just had. What do you think?

docs/source/examples/input_schemas.rst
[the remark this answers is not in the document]
jimPruyne: I would try not to differentiate those unless we have to. A single schema should do it in all the cases. Either way, outside the scope of this PR.

globus_action_provider_tools/data_types.py
```diff
@@ -105,18 +104,17 @@ class ActionRequest(BaseModel):
         None,
         description="A timestamp indicating by which time the action must complete. The request may be rejected if the Action Provider does not expect to be able to complete the action before the deadline or if it represents a time greater than the maximum_deadline specified in the Provider Description.",
     )
-    release_after: Optional[str] = Field(
+    release_after: Optional[datetime.timedelta] = Field(
```
jimPruyne, line 107: This is excellent. We have not been consistent on this, and pydantic seems to unify this whole thing. Most happy about this!!!

jimPruyne (approved):
All looks good to me now. Especially love the use of the timedelta in the model for this change and that is another big win of going the pydantic route here.

PR #26, opened 2021-03-10
Add support for INACTIVE action resume

jimPruyne:
This adds another method to the APT blueprint for doing a resume
operation. This is to be called to signal that an INACTIVE action
status may be resumed. This includes the case where the action is
INACTIVE because it requires additional consent which should be
signaled to the user by using the `ConsentRequired` value on the
`code` of the action status details returned and setting the
`required_scope` value to the scope the user should consent to
recover.

To help with setting proper details on `INACTIVE` state and on
`FAILED` states helper model classes have been added for encoding the
correct properties on FAILED and INACTIVE action statuses. These are
intended to be extensible by the user, but should include the `code`
and `description` fields. Further extending these for particular
cases (e.g. the ConsentRequired INACTIVE cause) could/should be
considered.

### DIFF
--- a/globus_action_provider_tools/data_types.py
+++ b/globus_action_provider_tools/data_types.py
@@ -1,27 +1,16 @@
 import datetime
 import inspect
-from enum import Enum, auto
+from enum import Enum
 from json import JSONEncoder
 from os import urandom
-from typing import (
-    AbstractSet,
-    Any,
-    Dict,
-    Iterable,
-    List,
-    NamedTuple,
-    Optional,
-    Set,
-    Type,
-    Union,
-)
+from typing import AbstractSet, Any, Dict, List, Optional, Set, Type, Union
 
 import arrow
 import isodate
 from base62 import encodebytes as base62
 from pydantic import BaseModel, Field
 
-from .authentication import AuthState
+from globus_action_provider_tools.authentication import AuthState
 
 
 def now_isoformat():
@@ -85,14 +74,30 @@ class ActionProviderDescription(BaseModel):
     event_types: Optional[List[EventType]] = None
 
 
+_uuid_regex = (
+    "([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})"
+)
+_principal_urn_regex = f"^urn:globus:(auth:identity|groups:id):{_uuid_regex}$"
+
+
 class ActionRequest(BaseModel):
     request_id: str = Field(
         ...,
-        description="A unique identifier representing the request to start an action. Multiple uses of the same request_id must have the same content or they will be rejected. Only one instance of the operation will be executed, so requests with the same request_id may be repeated to attempt to guarantee execution of an action",
+        description=(
+            "A unique identifier representing the request to start an action. "
+            "Multiple uses of the same request_id must have the same content or they "
+            "will be rejected. Only one instance of the operation will be "
+            "executed, so requests with the same request_id may be repeated "
+            "to attempt to guarantee execution of an action"
+        ),
     )
     body: Dict[str, Any] = Field(
         ...,
-        description="The Action Provider-specific content describing the action to run. The format for the body is provided in the input_schema field of the Action Provider Description",
+        description=(
+            "The Action Provider-specific content describing the action "
+            "to run. The format for the body is provided in the "
+            "input_schema field of the Action Provider Description"
+        ),
     )
     label: Optional[str] = Field(
         None,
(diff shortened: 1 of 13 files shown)


urimandujano:
@jimPruyne can you check up on why the CI is failing?

jimPruyne:
@urimandujano Looks like tests are passing now.

globus_action_provider_tools/data_types.py
```diff
@@ -85,14 +74,30 @@ class ActionProviderDescription(BaseModel):
     event_types: Optional[List[EventType]] = None
 
 
+_uuid_regex = (
+    "([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})"
+)
+_principal_urn_regex = f"^urn:globus:(auth:identity|groups:id):{_uuid_regex}$"
```
urimandujano, line 80: These are a good idea, thanks for this

globus_action_provider_tools/data_types.py
```diff
@@ -102,35 +107,111 @@ class ActionRequest(BaseModel):
...
+            return vars(self).pop(key, args[0])
+        else:
+            return vars(self).pop(key)
+
+
+class ExtensibleCodeDescription(BaseModel, SubscriptableObject):
+    class Config:
+        extra = "allow"
+
+    code: str = Field(..., description=(""))
```
urimandujano, line 177: These description fields are usually used when this model gets converted into JSONSchema, but in this case since they're internal models I don't think we need to set them to anything at all.

globus_action_provider_tools/data_types.py
```diff
@@ -102,35 +107,111 @@ class ActionRequest(BaseModel):
...
+    pass
+
+
+class PaginationWrapper(BaseModel):
+    limit: int
+    has_next_page: bool
+    marker: Optional[str]
+
+
+class ActionLogEntry(ExtensibleCodeDescription):
```
urimandujano, line 191: I guess we could default to the `GenericLogEntry` code here. That might make it a little easier to construct these, but it could be that you were thinking there'd be different types of log entries, in which case I think this is good.

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -247,6 +243,52 @@ def wrapper() -> ViewReturn:
         self.add_url_rule("/actions", func.__name__, wrapper, methods=["POST"])
         return wrapper
 
+    @overload
+    def action_resume(self, func: Callable[[str, AuthState], ActionStatusReturn]):
+        """
+        Using these stubs w/ @overload tells mypy that the actual implementation
+        for action_status can accept a str or ActionStatus as the first arg type
```
urimandujano, line 250: Can you update this to `action_resume`?

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -247,6 +243,52 @@ def wrapper() -> ViewReturn:
...
+        """
+        ...
+
+    @overload
+    def action_resume(
+        self, func: Callable[[ActionStatus, AuthState], ActionStatusReturn]
+    ):
+        """
+        Using these stubs w/ @overload tells mypy that the actual implementation
+        for action_status can accept a str or ActionStatus as the first arg type
```
urimandujano, line 262: Can you update this to `action_resume`?

urimandujano (commented):
I had a few minor comments and questions but on the whole I really like the direction this PR takes us and I'm excited to see what else this let's us do with ActionProviders.

If you can address some of my thoughts then I think this is good to go.

globus_action_provider_tools/data_types.py
jimPruyne: Yeah, I don't know about defaulting to that. On the one hand, we don't really specify what the codes should/could be. On the other hand, I kinda like people who are actually gonna do logs (probably a small subset) to think about what they want there.

globus_action_provider_tools/flask/apt_blueprint.py
jimPruyne: I don't know how I made this mistake after spending so much time thinking carefully exactly what I wanted to write here. :)

Good catch.

jimPruyne:
@urimandujano typos fixed. Ready for final(?) review.

globus_action_provider_tools/data_types.py
urimandujano: > I kinda like people who are actually gonna do logs

FTFY :)

urimandujano (approved):

PR #29, opened 2021-04-16
Improves error reporting via logs

urimandujano:
- Create app exception logs when the AP encounters and returns an Exception
- Validating input bodies throw uniform BadActionRequests
- Logs errors when validating input bodies
- Handles + logs errors where JSON input can't be parsed


jimPruyne (approved):
Any particular reason to change the typing imports?

urimandujano:
No reason, it's just a new style thing I've fallen into. I've enjoyed it so far since it makes using typing annotations more "grab bag" like.

PR #30, opened 2021-04-20
Fixes return type for action resume

urimandujano:
The resume endpoint was only handling return values which were `ActionStatus`, not a tuple of [`ActionStatus`, `returnCode`]. This update allows for the resume endpoint to handle either or. This makes it trivial for an AP implementer to return their own status code when the resume operation is called.


jimPruyne (approved):
This was the proper fix to what I patched in the transfer APs. Thanks for following up and getting it in place.

PR #31, opened 2021-04-22
Updates and stabilizations preparing for production use

urimandujano:
- Publishes the APT OpenAPI spec to GH pages
- Updates Makefile help and commands
- Bundles Flask as an optional dependency
  - When installing APT, only the core components are installed. To install and use the "Flask" pieces, install using extras: pip install globus_action_provider_tools[flask]
- Updates README with how and why to install the flask extras
- Updates README with information on making PyPi releases
- Cleanly separates the Flask library components from the toolkit (partiularly the Flask Exception classes)
- Stabilizes the package's API
- Adds more logging to the authorization helpers
- Adds shared utilities to public interface
- Adds deprecation warning to flask callback loader function


urimandujano:
@jimPruyne is this something you have bandwidth for? I think this PR in particular would be good to have in place before GW. I also anticipate some work to get these updates integrated on our AP side

CI is passing so if we'd rather just use that as a signal I can merge, deploy, and work out any kinks on updating our APs.

PR #45, opened 2021-12-09
Fix for missing or improperly cached refresh tokens

jimPruyne:
The previous implementation was checking for presence of the
`refresh_token` key in the dependent token response object. It turns
out that doing an `in` check always returns true, so this changes to
checking for the value of `None` rather than doing an `in` check. In the
previous mode, we could attempt to create a `RefreshTokenAuthorizer`
with a `None` `refresh_token` which will always fail. This, in turn,
would not show up until an AP attempted to use the authorizer via an
SDK client, like TransferClient, which could make it hard to trace
back to the root cause.

We also change the dependent token cache to a `TTLCache` with timeout
of 47 hours instead of a basic `LRUCache` so that we don't keep tokens
around longer than the expected expiration any way. Thus, the
expectation is that in the common case, prior to any tokens expiring
and attempting a refresh, we will, instead, no retain the tokens which
are due to expire and cause a new dependent grant. This is somewhat
non-optimal assuming dependent grants are more expensive, particularly
for Auth, than refresh operations, but it also makes handling missing
refresh tokens in some way automatic since we don't use them as often.

To that last, the source for a lot of this is that the dependent
grants do not seem to be returning refresh tokens at all in some
cases. That led to the whole path of building this additional
robustness that we didn't expect to need for missing refresh
tokens. This robustness here is the right thing, but why refresh
tokens are not coming in some cases is still an unsolved issue.

### DIFF
--- a/changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
+++ b/changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
@@ -22,8 +22,8 @@
 ..
 .. - A bullet item for the Deprecated category.
 ..
-.. Fixed
-.. -----
+Fixed
+-----
 ..
 .. - A bullet item for the Fixed category.
 ..
--- a/changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
+++ b/changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
@@ -1,0 +1,35 @@
+.. A new scriv changelog fragment.
+..
+.. Uncomment the header that is right (remove the leading dots).
+..
+.. Removed
+.. -------
+..
+.. - A bullet item for the Removed category.
+..
+.. Added
+.. -----
+..
+.. - A bullet item for the Added category.
+..
+.. Changed
+.. -------
+..
+.. - A bullet item for the Changed category.
+..
+.. Deprecated
+.. ----------
+..
+.. - A bullet item for the Deprecated category.
+..
+.. Fixed
+.. -----
+..
+.. - A bullet item for the Fixed category.
+..
+* Fixed handling of missing refresh tokens in dependent token grants. Now, even if a refresh token is expected in a dependent grant, it falls back to just using the access token up until the time the access token expires. We also shorten the dependent token grant cache to be less than the expected lifetime of an access token and, thus, from cache, we should not retrieve an access token which is already expired.
+.. Security
+.. --------
+..
+.. - A bullet item for the Security category.
+..
--- a/globus_action_provider_tools/authentication.py
+++ b/globus_action_provider_tools/authentication.py
@@ -210,7 +210,16 @@ class AuthState(object):
 
         refresh_token = dep_tkn_resp.get("refresh_token")
         access_token = dep_tkn_resp.get("access_token")
-        token_expiration = dep_tkn_resp.get("expires_at_seconds")
+        token_expiration = dep_tkn_resp.get("expires_at_seconds", 0)
+        # IF for some reason the token_expiration comes in a string, or even a string
+        # containing a float representation, try converting to a proper int. If the
+        # conversion is impossible, set expiration to 0 which should force some sort of
+        # refresh as described elsewhere.
+        if not isinstance(token_expiration, int):
+            try:
+                token_expiration = int(float(token_expiration))
+            except ValueError:
+                token_expiration = 0
         now = time()
 
         # IF we have an access token, we'll try building an authorizer from it if it is
--- a/globus_action_provider_tools/authentication.py
+++ b/globus_action_provider_tools/authentication.py
@@ -27,8 +27,9 @@ class AuthState(object):
     # Cache for introspection operations, max lifetime: 30 seconds
     introspect_cache: TTLCache = TTLCache(maxsize=100, ttl=30)
 
-    # Cache for dependent tokens, max lifetime: 5 minutes
-    dependent_tokens_cache: LRUCache = LRUCache(maxsize=100)
+    # Cache for dependent tokens, max lifetime: 47 hours: a bit less than the 48 hours
+    # until a refresh would be required anyway
+    dependent_tokens_cache: TTLCache = TTLCache(maxsize=100, ttl=47 * 3600)
 
     # Cache for group lookups, max lifetime: 5 minutes
     group_membership_cache: TTLCache = TTLCache(maxsize=100, ttl=60 * 5)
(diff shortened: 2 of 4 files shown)


changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
```diff
@@ -0,0 +1,35 @@
...
+.. ----------
+..
+.. - A bullet item for the Deprecated category.
+..
+.. Fixed
+.. -----
+..
+.. - A bullet item for the Fixed category.
+..
+* Fixed handling of missing refresh tokens in dependent token grants. Now, even if a refresh token is expected in a dependent grant, it falls back to just using the access token up until the time the access token expires. We also shorten the dependent token grant cache to be less than the expected lifetime of an access token and, thus, from cache, we should not retrieve an access token which is already expired.
```
kurtmckee, line 30: The `..`  prefix needs to be removed so it looks like:

```rst
Fixed
-----

* Fixed handling...
```

globus_action_provider_tools/authentication.py
```diff
@@ -177,15 +171,20 @@ def get_authorizer_for_scope(
...
                 self.auth_client,
-                access_token=dep_tkn_resp["access_token"],
-                expires_at=dep_tkn_resp["expires_at_seconds"],
+                access_token=access_token,
+                expires_at=token_expiration,
             )
-        elif "access_token" in dep_tkn_resp:
-            return AccessTokenAuthorizer(dep_tkn_resp["access_token"])
+        elif access_token is not None and token_expiration > now:
+            return AccessTokenAuthorizer(access_token)
```
kurtmckee, line 187: In practice it seems unlikely that you'll have a response that contains an `"access_token"` key but no `"token_expiration"` key. Nevertheless, this is not verifying the type of `token_expiration` and may raise a `TypeError` as a result.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "globus-action-provider-tools"
-version = "0.11.4"
+version = "0.11.5"
```
kurtmckee, line 3: I don't think the version should bump in this branch.

kurtmckee (changes requested):

globus_action_provider_tools/authentication.py
jimPruyne: Good one, I ended up putting some semi-convoluted conversion to int attempt

changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
jimPruyne: Thanks, scriv noobness showing through.

jimPruyne:
@kurtmckee Thanks for the review, updated as suggested.

jimPruyne:
@kurtmckee @urimandujano @rudyardrichter I had pushed one revision that cleaned things up, but I wasn't too satisfied with the behavior in all cases (particularly where we have an expired access token, but no refresh token).

So, I've updated the logic in deciding how / when to return a refresh authorizer vs attempting another dependent grant. The logic is more convoluted, but I hope that the robustness makes it worth it.

Sorry to make such an extensive change after requesting review.

globus_action_provider_tools/authentication.py
```diff
@@ -176,17 +213,37 @@ def get_authorizer_for_scope(
         token_expiration = dep_tkn_resp.get("expires_at_seconds")
```
kurtmckee: `token_expiration` has the possibility to be `None`, and can still raise a `TypeError` in the code below when compared to `now`. Would it make sense to use a default of `0` so that it will be the correct type and can still fail the inequality comparison correctly?

```python
token_expiration = dep_tkn_resp.get("expires_at_seconds", 0)
```

globus_action_provider_tools/authentication.py
```diff
@@ -176,17 +213,37 @@ def get_authorizer_for_scope(
...
         now = time()
```
kurtmckee, line 223: `now` is only used in one place. Would it be worthwhile to remove this variable and get the current `time()` in the one place where it's used?

kurtmckee (changes requested):

changelog.d/20211209_160927_jimPruyne_mishandled_refresh_token_missing_on_dependent_grant.rst
kurtmckee: This is strange, I don't have a "resolve" button for this. Thanks for the update, Jim!

pyproject.toml
kurtmckee: Again, no resolve button but this is fixed. Thanks Jim!

globus_action_provider_tools/authentication.py
kurtmckee: I was looking at an out-of-date version. I'm deleting this comment. Sorry for the confusion!

globus_action_provider_tools/authentication.py
```diff
@@ -152,42 +146,113 @@ def dependent_tokens_cache_id(self) -> Optional[str]:
...
+        refresh_token = dep_tkn_resp.get("refresh_token")
+        access_token = dep_tkn_resp.get("access_token")
+        token_expiration = dep_tkn_resp.get("expires_at_seconds", 0)
+        # IF for some reason the token_expiration comes in a string, or even a string
+        # containing a float representation, try converting to a proper int. If the
+        # conversion is impossible, set expiration to 0 which should force some sort of
+        # refresh as described elsewhere.
+        if not isinstance(token_expiration, int):
+            try:
+                token_expiration = int(float(token_expiration))
```
kurtmckee, line 220: A bit of trivia (that will never happen within the context of our services): it's possible for this code to raise an `OverflowError`, too!

```python
int(float("inf"))  # 💥 OverflowError: cannot convert float infinity to integer
```

Within the context of our API's this will never happen but it's possible when interacting with untrusted sources.

kurtmckee (approved):

urimandujano (approved):
This looks good to me

rudyardrichter (approved):

kurtmckee (approved):

urimandujano:
@jimPruyne I was thinking about this this afternoon and trying things out and I learned that the `dependent_token_cache_id` stays constant across different access tokens. So even if we send fresh tokens, its `dependent_token_cache_id` would be the same, which would cause a cache hit, which would get us an old dependent token grant response. Using the old `LRUCache` meant that we were never updating our dependent tokens. I feel pretty sure that switching to a `TTLCache` like you've done here will make things better!

I'm building on this PR with one that adds a bit more logging around cache activities.

rudyardrichter:
A squash prior to merging would be excellent

PR #55, opened 2022-02-02
Upgrade to globus-sdk v3

rudyardrichter:
- Add note on upgrade to changelog.
- Update GroupsClient to work with v3.
    - Update groups client usage
    - Add basic test for creating groups client (similar to flows)
- Remove `arrow` as a dependency
    - There was only one usage site for this, which is also easily accomplished using the standard library, so it seems good to trim the dependency here


pyproject.toml
```diff
@@ -31,10 +31,9 @@ whattimeisit-provider = "examples.flask.whattimeisitrightnow.app.app:main"
 
 [tool.poetry.dependencies]
 python = "^3.6.2"
-globus-sdk="^2"
+globus-sdk="^3.3.0"
 jsonschema = "^3"
 pyyaml = "^5.3.1"
-arrow = "^1"
```
kurtmckee, line 37: Glad to see this dependency go since it only appears to be used in one place.

globus_action_provider_tools/utils.py
```diff
@@ -1,11 +1,11 @@
+import datetime
 from os import urandom
 
-import arrow
 from base62 import encodebytes as base62
 
 
 def now_isoformat():
-    return str(arrow.utcnow())
+    str(datetime.datetime.now(datetime.timezone.utc).isoformat())
```
kurtmckee, line 8: The missing `return` statement suggests that this function might not be in use (at least, not during unit testing).

kurtmckee (changes requested):

globus_action_provider_tools/utils.py
rudyardrichter: Aha, I have successfully discovered the need for better unit test coverage/type annotations...

rudyardrichter:
@kurtmckee really good catch, thanks. That prompted me to root around the tests a bit; it was possible to (IMO) simplify the test setup quite a bit, and in my case xdist was actually _slowing_ the test run (it took 3s vs 1.5 without) so I'm removing that too.

kurtmckee (approved):

PR #56, opened 2022-02-11
Release/0.12.0b1

kurtmckee:
# Changelog

Features
--------

- Upgrade to use major version 3 of the [Globus SDK](https://github.com/globus/globus-sdk-python). If you are using Action
  Provider Tools in an environment which is currently using an earlier version
  of the Globus SDK, then you will need to upgrade first in order for this
  version to be compatible.

Bugfixes
--------

- Fixes an issue where the `ActionProviderBlueprint` decorators were not
  returning the decorated functions. This meant that the registered functions
  were loaded onto the Action Provider correctly but were `None` in the module
  in which they were defined.


CHANGELOG.rst
```diff
@@ -10,6 +10,31 @@ Unreleased changes are documented in files in the `changelog.d`_ directory.
...
+- Fixes an issue where the `ActionProviderBlueprint` decorators were not
+  returning the decorated functions. This meant that the registered functions
+  were loaded onto the Action Provider correctly but were `None` in the module
+  in which they were defined.
+
+Documentation
+-------------
+
+- Document the git workflow in ``WORKFLOW.md``.
+
```
rudyardrichter, line 37: This point is not user-facing.

rudyardrichter (commented):
one comment on changelog, otherwise LGTM

CHANGELOG.rst
kurtmckee: Removed.

PR #61, opened 2022-06-22
Remove the `__version__` metadata attribute

kurtmckee:
Changes
-------

-   Remove the ``__version__`` attribute.

    The ``importlib.metadata`` module in Python 3.8 and higher
    (or the backported ``importlib_metadata`` package)
    can be used to query the version of installed packages if needed.


joshbryan-globus (approved):
I'm fine removing this, though it is also referenced in [this comment](https://github.com/globus/action-provider-tools/blob/main/docs/source/conf.py#L28)

kurtmckee:
Thanks for the quick review, Josh! I've removed that reference, and while confirming the docs could build...discovered that they can't with Sphinx 3.2 and Python 3.10. I upgraded the Sphinx dependency to the latest, which allows the docs to build.

I also updated the docs copyright.

PR #68, opened 2022-10-26
Address invalid grant errors

kurtmckee:
> NOTE:
>
> This PR temporarily targets a non-main branch as its base to isolate the commit this fix.
>
> The base branch can be switched back to `main` when PR #65 is merged.

Bugfixes
--------

- Fix a crash that occurs when an HTTP 400 "invalid grant" error is received
  from Globus Auth while getting an authorizer for a given scope.

  This is now caught by ``AuthState.get_authorizer_for_scope()`` and ``None`` is returned.

### DIFF
--- a/changelog.d/20221026_165824_kurtmckee_address_invalid_grant_error_sc_19538.rst
+++ b/changelog.d/20221026_165824_kurtmckee_address_invalid_grant_error_sc_19538.rst
@@ -1,0 +1,7 @@
+Bugfixes
+--------
+
+- Fix a crash that occurs when an HTTP 400 "invalid grant" error is received
+  from Globus Auth while getting an authorizer for a given scope.
+
+  This is now caught by ``AuthState.get_authorizer_for_scope()`` and ``None`` is returned.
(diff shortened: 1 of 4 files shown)


derek-globus (approved):
Not sure I agree with the underlying interface of this method but the change is inline with the established interface so I'm fine approving it as bugfix.

The interface seems to be:
"Do your best effort at getting an Authorizer for me. If you can't, don't tell me (the caller) why you couldn't; rather emit a `warning` because you as a re-usable method don't have the context that I do to know whether not being able to get an Authorizer is cause for concern."

I think a much better interface is:
"Try to get me an authorizer, if you can't throw an error explaining why you couldn't. I as the caller have the context to know whether I should (1) try with a different scope, (2) let the error bubble up and fail the request, (3) translate it into a more meaningful error message (based on the type of error you threw)."

kurtmckee:
I completely agree with your assessment. As a bugfix, this seems okay. As a design, I think things should raise documented exceptions and callers should use that information to make decisions.

kurtmckee:
_Rebased on `main`, resolved merge conflicts, and force-pushed_

PR #69, opened 2022-10-28
Patch [sc-19668]: Rephrase JsonProvider to avoid type error

sirosen:
I'm okay with not changing this, but this is my thought on how to avoid the type-ignore.

I think `mypy` sees `class` as redefining a name in a way that it does not see assignment. It's a type-checker artifact, and I could understand the complaint that the classname now has a misleading underscore, but I think this is still nicer than the type-ignore.

### DIFF
--- a/globus_action_provider_tools/flask/helpers.py
+++ b/globus_action_provider_tools/flask/helpers.py
@@ -4,7 +4,7 @@ import inspect
 import json
 from enum import Enum
 from functools import partial
-from typing import Any, Callable, Dict, Iterable, Optional, Set, Type
+from typing import Any, Callable, Dict, Iterable, Set, Type
 
 import flask
 from flask import Request, current_app, jsonify
@@ -212,15 +212,18 @@ try:
 except ImportError:
     # Flask < 2.2: Use the deprecated JSON encoder interface.
     json_provider_available = False
-    JsonProvider: Optional["DefaultJSONProvider"] = None
+
+    JsonProvider: type[DefaultJSONProvider] | None = None
 else:
     # Flask >= 2.2: Use the new JSON provider interface.
     json_provider_available = True
 
-    class JsonProvider(DefaultJSONProvider):  # type: ignore[no-redef]
+    class _JsonProvider(DefaultJSONProvider):
         @staticmethod
         def default(o: Any) -> Any:
             return convert_to_json(o)
+
+    JsonProvider = _JsonProvider
 
 
 def assign_json_provider(app_or_blueprint: flask.Flask | flask.Blueprint):


globus_action_provider_tools/flask/helpers.py
```diff
@@ -212,16 +212,19 @@ def pydantic_input_validation(
 except ImportError:
     # Flask < 2.2: Use the deprecated JSON encoder interface.
     json_provider_available = False
-    JsonProvider: Optional["DefaultJSONProvider"] = None
+
+    JsonProvider: type[DefaultJSONProvider] | None = None
```
kurtmckee, line 216: Wut. How does `type[DefaultJSONProvider]` not trigger a NameError? I obviously have more to learn...

I ran some quick test code and discovered that the following code WILL raise NameError _unless `from __future__ import annotations` is present.

```python
try:
    from bogus import nonsense
except ImportError:
    Demo: type[nonsense] | None = None
else:
    Demo = "the real thing goes here"
```

That's true going all the way back to Python 3.7 (3.7.13, at least, on my machine).

globus_action_provider_tools/flask/helpers.py
```diff
@@ -212,16 +212,19 @@ def pydantic_input_validation(
...
     json_provider_available = True
 
-    class JsonProvider(DefaultJSONProvider):  # type: ignore[no-redef]
+    class _JsonProvider(DefaultJSONProvider):
         @staticmethod
         def default(o: Any) -> Any:
             return convert_to_json(o)
 
+    JsonProvider = _JsonProvider
+
```
kurtmckee, line 227: I'm astonished this works, but if we take this route we've merely dodged leaving a comment that explicitly states "mypy, ignore this well-defined error" but would then have to add a comment stating "human, this strange incantation pleases the current implementation of our most handsome automatons".

kurtmckee (commented):
I'm not satisfied by this alternative. I'd like to keep the explicit "no-redef" version, which is self-explanatory by its very nature.

globus_action_provider_tools/flask/helpers.py
kurtmckee: A more basic example:

```python
instance: NoProblemsHere  # NameError!

from __future__ import annotations

instance: NoProblemsHere  # And indeed there are no problems!
```

sirosen:
> I'm not satisfied by this alternative. I'd like to keep the explicit "no-redef" version, which is self-explanatory by its very nature.

I think that's reasonable. I thought maybe I could make something cleaner than this, but this was the "best" variation I could come up with. I figured it was worth sharing before deciding on my own.

I'm closing and will approve #65, but I'll also come back here to see if we should discuss the comments at all. 🙂

globus_action_provider_tools/flask/helpers.py
sirosen: Yeah, this is `__future__` magic.

I'm not sure if you've followed some of the threads about annotations maybe becoming a data descriptor, but the implications are odd. With deferred annotations pulled from 3.11, we might be in an unprecedented situation in which there is a future import _which will never become the future behavior_. I may need to rethink how aggressively I lean on the future import behavior in my code.

globus_action_provider_tools/flask/helpers.py
sirosen: Most handsome _and_ best dressed. 😉 

I agree that it's a lot of handwaving just to make the ignore comment go away.
To be perfectly honest, I was disappointed that I couldn't do better than this. I thought that maybe if I played with the order of the definitions I could make the error go away in a more "natural" way, but nothing else worked.

globus_action_provider_tools/flask/helpers.py
kurtmckee: > there is a future import which will never become the future behavior

That's the exact plot of "Terminator: Dark Fate"! Is James Cameron a Python release manager?

kurtmckee: Jokes aside, this is all very surprising. I don't have an opinion about what _should_ happen in the future, only that this broke my expectations.

PR #71, opened 2022-12-01
Update to jsonschema version '^4.17'

sirosen:
- Update to the latest jsonschema versions.
- Replace all use of Draft7Validator with `validator_for` to support arbitrary drafts

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -8,36 +8,36 @@ python-versions = "*"
 
 [[package]]
 name = "attrs"
-version = "21.4.0"
+version = "22.1.0"
 description = "Classes Without Boilerplate"
 category = "main"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+python-versions = ">=3.5"
 
 [package.extras]
-dev = ["cloudpickle", "coverage[toml] (>=5.0.2)", "furo", "hypothesis", "mypy", "pre-commit", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six", "sphinx", "sphinx-notfound-page", "zope.interface"]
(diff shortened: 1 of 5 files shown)


ada-globus (approved):
Needs a changelog entry, but otherwise LGTM!

sirosen:
What do you mean? It has a changelog fragment! 😝

ada-globus:
> What do you mean? It has a changelog fragment! 😝

Retcon FTW ;)

poetry.lock
```diff
@@ -8,36 +8,36 @@ python-versions = "*"
 
 [[package]]
 name = "attrs"
-version = "21.4.0"
+version = "22.1.0"
```
kurtmckee, line 11: I didn't realize we were keeping a `poetry.lock` file in-repo. For a library this surprises me....

kurtmckee (commented):
Should we delete `poetry.lock`?

sirosen:
> Should we delete `poetry.lock`?

I don't really know. I'm very unclear on what the best practices are for library development with `poetry`.
The only purpose I can think of for `poetry.lock` when developing a library is to record a set of dependency versions which was tested and known to work.

However, if we're removing `poetry.lock`, I might suggest switching to `setuptools` + `build` + `twine` (like `globus-sdk`). Without the lockfile, why use `poetry`? I'm not sure that the value proposition for the tool is very strong without `poetry.lock`.

kurtmckee:
The [poetry docs state](https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control) that:

> For libraries it is not necessary to commit the lock file.

That's why I'm surprised to find that `poetry.lock` was committed to this repo.

As a value add, poetry shines for multiple reasons across several categories, including:

* Automatic trove classifiers for license and supported Python versions during packaging; automatic README injection; etc
* A singular CLI tool with bash and zsh command completion
* Uses a declarative configuration file instead of running Python code in `setup.py`
* Integrated virtual environment creation and entry
* Automatic installation of extras (without having to open a file to find out what the extras are)
* Consistency of tool use within most of the team's repos

Some of these reasons could apply to other tools instead of poetry, but I don't think they apply to setuptools, build, and twine, which I consider a step backwards from what poetry (and its like) offers.

PR #74, opened 2022-12-14
Release/0.13.0b1

kurtmckee:
Python support
--------------

- Add support for Python 3.11.
- Drop support for Python 3.6.

Bugfixes
--------

-   Fix a crash that will occur if a non-object JSON document is submitted.
    For example, this will happen if the incoming JSON document is ``"string"``
    or ``["array"]``.

- Fix a crash that occurs when an HTTP 400 "invalid grant" error is received
  from Globus Auth while getting an authorizer for a given scope.

  This is now caught by ``AuthState.get_authorizer_for_scope()`` and ``None`` is returned.

Changes
-------

-   Remove the ``__version__`` attribute.

    The ``importlib.metadata`` module in Python 3.8 and higher
    (or the backported ``importlib_metadata`` package)
    can be used to query the version of installed packages if needed.

- ``jsonschema>=4.17,<5`` is now required by action-provider-tools.

  Consumers of the library will have to update to a more recent version of ``jsonschema``
  if they are using it explicitly.


sirosen:
We agreed previously that this repo should not have a `poetry.lock` file in VCS. Should we scrub from release notes all of the changelog items which refer to changes to that file?
For example,
```
- Update the `pyjwt` package to version 2.4.0.
  This mitigates a security vulnerability.
```
is misleading. We don't have a direct dependency on a pinned version of `pyjwt` here.

kurtmckee:
Good point. I'll update this.

kurtmckee:
`poetry.lock` has been removed and git-ignored, and the CHANGELOG has been updated and reorganized to remove references to dev dependencies and under-the-hood unit testing changes.

sirosen (approved):

PR #77, opened 2022-12-16
Remove a no-op call to Globus Auth during ``TokenChecker`` instantiation

kurtmckee:
Changes
-------

-   Remove an unused parameter from ``TokenChecker``: ``cache_config``.
-   Remove a no-op call to Globus Auth during ``TokenChecker`` instantiation.


sirosen (approved):
We discussed this in chat, but to clue others in:
The rationale behind the call to Auth is questionable. I am not convinced it is a good idea and believe that the addition was never reviewed or discussed by a team.

derek-globus (approved):

PR #78, opened 2022-12-16
Remove the "deprecated" module and cascade

sirosen:
Opened as a draft to start the conversation.
My own opinion is that we should remove these examples until the library is more stable, but I'm open to discussing further.

---

- remove '..flask.api_helpers'
- remove all test fixtures which depend on this module
  - modify tests only as necessary to handle the removal
- remove examples which depend on this module


sirosen:
Closing in favor of #81. I think this was useful to clarity the impact of the potential removal, but I was not confident we should make the change in the first place.

PR #83, opened 2023-01-19
Overhaul testing

kurtmckee:
# Testing performance improvements

Times below are rough and may vary, particularly in CI testing.

## Local tox runs

These changes result in a drop from 20 seconds to 3 seconds when running tests locally and in parallel.

```
           +------------+------------+
           |   Before   |    After   |
+----------+------------+------------+
| Serial   | 20 seconds | 12 seconds |
| Parallel |  --n/a--   |  3 seconds |
+----------+------------+------------+
```

## CI runs

When iterating on a branch, assuming a warm cache, these changes result in a CI usage drop from 9m12s to only 4m45s. The wall clock time (total duration) also decreases.

```
                     +---------+-------+
                     | Before  | After |
+--------------------+---------+-------+
| Cache miss (total) | 3m19s   | 2m31s |
| Cache miss (usage) | 9m12s   | 9m32s |
| Cache hit  (total) | --n/a-- | 2m16s |
| Cache hit  (usage) | --n/a-- | 4m45s |
+--------------------+---------+-------+
```

----

## Development

-   During local testing, build a shared wheel.

    Previously, a shared ``.tar.gz`` file was created.
    However, in each tox environment, pip would convert this to a wheel during installation.

    This change decreases local test times from ~20 seconds to ~12 seconds.

-   Support running tox test environments in parallel (run ``tox p``).

    This change decreases local test times to only ~3 seconds.

-   Overhaul CI.

    -   Introduce caching of the ``.tox/`` and ``.venv/`` directories.

        The cache is invalidated once each week (``date %U`` rolls the week on Sundays).

    -   Build a shared wheel once as an artifact and re-use it across all test environments.
    -   Consolidate standard testing and testing of minimum Flask versions.


sirosen (approved):
Looks fabulous.

I think we may want to think again about reusable workflows -- it would be nice to be able to write this once, correctly, and then parametrize it over supported pythons, tox environment names, and perhaps one or two other variables.
(Or we could write a workflow file generator! 😉 )

kurtmckee:
> we could write a workflow file generator!

I'll add Jinja2 as a dev dependency and start coding this up, if you'll write the pre-commit hook to ensure it outputs valid YAML. kthx :stuck_out_tongue_winking_eye:

PR #98, opened 2023-09-15
Drop Python 3.7 support

kurtmckee:
_This also escalates test suite warnings to errors._

Python support
--------------

*   Drop support for Python 3.7.


sirosen (approved):
Yep, given the limited audience for this library I see no reason to go slow.
We should discuss dropping 3.7 from globus-sdk as well, where we have a broader userbase and support context.

PR #102, opened 2023-10-03
Support Python 3.12; update min. Flask; fix test suite

kurtmckee:
This should resolve issues in the test suite.


sirosen (approved):
I'll see if `freezegun` already has a fix in the works. Otherwise, maybe we can throw a PR their way.

PR #105, opened 2023-10-13
CloudWatch EMF Route RequestLifecycleHooks

derek-globus:
## What?
1. Added support for RequestLifecycleHook registration to the flask ActionProviderBlueprint class.
  Classes may be provided at Blueprint instantiation time to register before, after, or
  teardown functionality to wrap all view invocation.

2. Added a CloudWatchEMFLogger RequestLifecycleHook class.
  When attached to an ActionProviderBlueprint, it will emit request count, latency, and
  response category (2xxs, 4xxs, 5xxs) count metrics through CloudWatch EMF. Metrics
  are emitted both at the aggregate AP and at the individual route level.

Dimension Sets
* {ActionProvider: `<action-provider-name>`}
* {ActionProvider: `<action-provider-name>`, Route: `<route-type>`}
   * `<route-type>` in this case is provided by the library not the user view func, so it'll be "run", "release", "resume", etc.

Metrics:
* Count - Count of invocations against an AP or route
* 2XXs - Count of invocations against an AP or route resulting in a 2XX
* 4XXs - Count of invocations against an AP or route resulting in a 4XX
* 5XXs - Count of invocations against an AP or route resulting in a 5XX
* RequestLatency - Duration of invocation handling for an AP or route (in milliseconds)

## Why?
Provide an easy mechanism for Action Providers to get usage metrics for both their aggregate Action Provider & specific routes of action providers

## Usage
```python
from globus_action_provider_tools.flask import ActionProviderBlueprint
from globus_action_provider_tools.flask.request_lifecycle_hooks import (
    CloudWatchMetricEMFLogger,
)

aptb = ActionProviderBlueprint(
    name="TrackedActionProvider",
    import_name=__name__,
    url_prefix="/tracked",
    provider_description=ap_description,
    request_lifecycle_hooks=[
        CloudWatchMetricEMFLogger(namespace="ActionProviders", action_provider_name="TrackedActionProvider")
    ],
)
```

## Testing
unit tested


sirosen (commented):
I know I'm jumping into a naming debate here, but I'd like us to avoid using "middleware" for this, specifically because Flask and Werkzeug already have a type of object they support called a "middleware".
In their case, it's something which wraps a whole app, e.g.
```python
app = app_factory()
app = CORSMiddleware(app, cors_config)
```

So I'm trying to find an alternative name which means basically the same thing.
"Lifecycle Hooks" or something like that?
Observer? (Not quite right, the hooks can modify data...)
HookCollection?

I'm struggling for a good name, but want to avoid an already-taken framework term.

derek-globus:
>I know I'm jumping into a naming debate here, but I'd like us to avoid using "middleware" for this, specifically because Flask and Werkzeug already have a type of object they support called a "middleware".

Yeah I didn't expect that I'd get this one through undebated 😬.

My reasoning for sticking with Middleware was that while flask does reference the term middleware, their "support" for it isn't really support. It's "hey wrap our app with a class that can do stuff and make sure it calls the app as well; we'll conventionally call that thing Middleware!" ([ref](https://flask.palletsprojects.com/en/2.3.x/quickstart/#hooking-in-wsgi-middleware))

I believe the interface I've exposed (a class that taps into their builtin `before_request`, `after_request`, `teardown_request` hooks) actually fits more precisely with the term middleware.

>So I'm trying to find an alternative name which means basically the same thing.
"Lifecycle Hooks" or something like that?
Observer? (Not quite right, the hooks can modify data...)
HookCollection?


`RequestLifecycleHooks` or `APRequestLifeCycleHooks` would probably be the only one that makes sense in that vein. Middleware is a more accurate naming, but to your point it may make sense to avoid any hint of conflict w/ Flask.

tests/test_flask_helpers/app_utils.py
```diff
@@ -67,7 +67,7 @@ class Config:
 )
 
 
-def test_action_enumeration(
+def mock_action_enumeration_func(
```
derek-globus, line 70: I renamed these functions because they were named with `test_` which caused pytest to think they were tests to be run when in fact they are utility callables for other tests to supply to the action provider blueprint as action provider view functions.

Example failure: https://github.com/globus/action-provider-tools/actions/runs/6510610219/job/17684687247?pr=105

ada-globus:
Agree with all of the above—"middleware" feels like the clearest, but the potential for confusion is high in this context. I like `RequestLifecycleHooks` quite a bit.

derek-globus:
I ran this PR by Stephen Zimmerman & he had the same comment about `Middlware` as terminology but was content with `RequestLifecycleHooks`.

Updating in new rev

sirosen:
> My reasoning for sticking with Middleware was that while flask does reference the term middleware, their "support" for it isn't really support. It's "hey wrap our app with a class that can do stuff and make sure it calls the app as well; we'll conventionally call that thing Middleware!" ([ref](https://flask.palletsprojects.com/en/2.3.x/quickstart/#hooking-in-wsgi-middleware))
> 
> I believe the interface I've exposed (a class that taps into their builtin `before_request`, `after_request`, `teardown_request` hooks) actually fits more precisely with the term middleware.

I agree with you, but I think the potential for confusion is real enough (e.g. me in 6 months) that we should use a slightly less optimal name because the best name was (rudely!) taken from us.

I'm on the `RequestLifecycleHooks` bandwagon now.

I'll circle back to do a real content-ful review later, but I think it's all good from a quick scan.
I scoffed at the `fmt: off` for a tick until I saw what it was wrapping, and felt like "okay, yeah, maybe we don't need to `black`-format _everything_". :wink: I'm sure you had a similar judgement which led you to turning it off.

derek-globus:
>I scoffed at the fmt: off for a tick until I saw what it was wrapping, and felt like "okay, yeah, maybe we don't need to black-format everything". 😉 I'm sure you had a similar judgement which led you to turning it off.

*cough cough* [flows-logging](https://github.com/globusonline/globus-flows/blob/main/src/globus_flows/logging/emf.py#L40-L49) *cough cough*

tests/test_flask_helpers/request_lifecycle_hooks/test_cloudwatch_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+    app.test_client().post("/tracked/run", json=req)
+    out, _ = capsys.readouterr()
+    emf_log = json.loads(out)
+
+    assert "_aws" in emf_log
+    assert emf_log.get("Count") == 1
+    assert emf_log.get("2XXs") == expected_2xxs
+    assert emf_log.get("4XXs") == expected_4xxs
+    assert emf_log.get("5XXs") == expected_5xxs
+    assert type(emf_log.get("RequestLatency")) == float
```
derek-globus, line 75: For some reason on py3.12 windows this request latency is 0.0 (my initial condition was request_latency > 0).

I don't know why that was the case (maybe it's inheriting freeze_time from other tests) but felt comfortable loosening the test to just verify that this is a number rather than a specific number/number range.


Example failure: https://github.com/globus/action-provider-tools/actions/runs/6510955778/job/17685743540

globus_action_provider_tools/flask/middleware/cloudwatch_metrics.py
```diff
@@ -0,0 +1,239 @@
...
+        # If a request errors mid-handling, after_request handlers will not be called,
+        #   so we need to emit metrics for errors separately here
+        if error:
+            if hasattr(g, "route_type") and hasattr(g, "request_start_time"):
+                status_code = 500
+                if hasattr(error, "code"):
+                    status_code = error.code
+                self.emit_route_metrics(
+                    route_name=g.route_type,
+                    request_latency=datetime.now() - g.request_start_time,
```
ada-globus: This may not matter terribly much in practice, but would it be better to use something like `time.perf_counter` or `time.monotonic` that is designed for this type of usage?

tests/test_flask_helpers/request_lifecycle_hooks/test_cloudwatch_metrics.py
ada-globus: Still reviewing, but I have a comment that _might_ be relevant to this, so I will push that before finishing review.

globus_action_provider_tools/flask/request_lifecycle_hooks/cloudwatch_metrics.py
```diff
@@ -0,0 +1,239 @@
...
+                "Dimensions": [["Foo"], ["Foo", "Bar"]],
+                "Metrics": [{ "Name": "MyCoolMetric", "Unit": "Milliseconds" }]
+            }]
+        },
+        "Foo": "a",
+        "Bar": "b",
+        "MyCoolMetric": 37,
+    }
+    ```
+    Note how there are two additional top-level keys besides "_aws".
```
ada-globus, line 152:
```suggestion
    Note how there are three additional top-level keys besides "_aws".
```

ada-globus (approved):
Looks great! Approving now as I don't have any substantive concerns, but have a couple minor notes for consideration.

globus_action_provider_tools/flask/request_lifecycle_hooks/cloudwatch_metrics.py
derek-globus: touché

globus_action_provider_tools/flask/middleware/cloudwatch_metrics.py
derek-globus: Sure I can switch; I'm not super familiar with the distinction between various ways of figuring out the current time in python.

derek-globus: Seems like `perf_counter` is the most designed for this use case, so switching to that.

derek-globus: This seems to have solved the problem.

Probably was running fast enough that something about integer conversion in datetime.now() & timedelta was truncating the sub-ms runtime in certain circumstances to 0.

Now it's emitted as a floating point value instead & CI run verifies it's > 0. Thanks for the recommendation.

tests/test_flask_helpers/request_lifecycle_hooks/test_cloudwatch_metrics.py
kurtmckee: I'll investigate on Windows with Python 3.12.

PR #113, opened 2023-10-27
Cache dependent tokens based on token hash

sirosen:


ada-globus (approved):
Looks great, and reviewed together in huddle!

globus_action_provider_tools/authentication.py
```diff
@@ -164,24 +170,20 @@ def groups(self) -> frozenset[str]:
...
     def get_dependent_tokens(self, bypass_cache_lookup=False) -> OAuthTokenResponse:
         """
         Returns OAuthTokenResponse representing the dependent tokens associated
         with a particular access token.
         """
-        dependent_tokens_cache_id = self.dependent_tokens_cache_id
-        if dependent_tokens_cache_id is not None and not bypass_cache_lookup:
-            resp = AuthState.dependent_tokens_cache.get(dependent_tokens_cache_id)
+        # Caching is done based on a hash of the token string, **not** the
+        # dependent_tokens_cache_id. This is intentional.
```
kurtmckee, line 179: > This is intentional

Why?

globus_action_provider_tools/authentication.py
```diff
@@ -190,11 +192,10 @@ def get_dependent_tokens(self, bypass_cache_lookup=False) -> OAuthTokenResponse:
         resp = self.auth_client.oauth2_get_dependent_tokens(
             self.bearer_token, additional_params={"access_type": "offline"}
         )
-        if resp is not None and dependent_tokens_cache_id is not None:
-            log.info(
-                f"Caching dependent token response for token ***{self.sanitized_token}"
-            )
-            AuthState.dependent_tokens_cache[dependent_tokens_cache_id] = resp
+        log.info(
+            f"Caching dependent token response for token ***{self.sanitized_token}"
```
kurtmckee, line 196: Would it be worthwhile to incorporate the `"***"` prefix into the `sanitized_token` value directly?

globus_action_provider_tools/authentication.py
sirosen: `dependent_token_cache_id` is only correct to use in narrow use cases because it is sometimes necessary to explicitly get and send a new token to the upstream service in order to get correct behavior.

The issue pertains to session data, which is retrieved via token introspection. Session data has no viable cache key, which leads to some services caching that data by hash of the access_token (the same strategy used here).
Dependent token caching using the cache ID from Auth intentionally indicates that new dependent tokens are *not* needed. This leads to stale session cache data in the upstream service.

Because any new session authentication flow (e.g. `globus session update --all`) produces new access tokens, session data caching based on access token hashing is _correct_ if the services only accepts calls directly from the client application. It's _incorrect_ when the service is sitting behind a proxy application like an action provider.

kurtmckee: :point_up: This explanation may be beneficial in-code, particularly for the benefit of other users / developers.

derek-globus: > ☝️ This explanation may be beneficial in-code, particularly for the benefit of other users / developers.


+1; if I saw us using a hash of the access token as a cache id rather than the field returned from auth called "dependent_token_cache_id", I'd think it was a bug.

sirosen: Kurt and I met to refine the way this is described to get at the heart of the matter without getting into trying to explain the multi-service state and so forth. Basically "using the access token hash results in new tokens when the AP gets new tokens, and therefore helps cache bust upstream".

globus_action_provider_tools/authentication.py
sirosen: Maybe? But I'd rather just get rid of that suffix-of-token-as-"sanitized" stuff.

A *very* early version of the SDK did this for its debug-level logging, and it seems to have accidentally promulgated the idea that a suffix of the token is a good value to use in logs. A hash is probably better for multiple reasons, and we could maybe even use the same hashing algo as Auth so that it matches the `at_hash` claim of the `id_token`, which might give us some nice symmetry.

kurtmckee (approved):

PR #130, opened 2024-03-19
Prevent `TypeError`s during pydantic location formatting

kurtmckee:
Bugfixes
--------

*   Prevent ``TypeError``s from occurring during pydantic error formatting.

    This was caused by integer list indexes in pydantic error locations.

### DIFF
--- a/globus_action_provider_tools/flask/helpers.py
+++ b/globus_action_provider_tools/flask/helpers.py
@@ -211,7 +211,16 @@ def pydantic_input_validation(
     try:
         validator(**action_input)
     except ValidationError as ve:
-        messages = [f"Field '{'.'.join(e['loc'])}': {e['msg']}" for e in ve.errors()]
+        messages = []
+        for error in ve.errors():
+            path = []
+            for location in error["loc"]:
+                if isinstance(location, str):
+                    path.append(f".{location}")
+                else:
+                    path.append(f"[{location}]")
+            field = "".join(path)[1:]  # Remove the leading period.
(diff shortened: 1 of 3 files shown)


sirosen (approved):

globus_action_provider_tools/flask/helpers.py
```diff
@@ -211,7 +211,16 @@ def pydantic_input_validation(
...
-        messages = [f"Field '{'.'.join(e['loc'])}': {e['msg']}" for e in ve.errors()]
+        messages = []
+        for error in ve.errors():
+            path = []
+            for location in error["loc"]:
+                if isinstance(location, str):
+                    path.append(f".{location}")
+                else:
+                    path.append(f"[{location}]")
+            field = "".join(path)[1:]  # Remove the leading period.
```
derek-globus, line 222: https://github.com/globusonline/globus-flows/blob/main/src/globus_flows/api/routes/blueprint/pydantic_errors.py#L31-L44

image

derek-globus (approved):

globus_action_provider_tools/flask/helpers.py
derek-globus: Technically we do it slightly differently in flows I guess.

* APs: `foo[0].bar`
* Flows: `$.foo[0].bar`

kurtmckee: I _knew_ I'd seen this brilliant formatting code somewhere before! :heart:

PR #131, opened 2024-03-22
Fix typos

kurtmckee:
The typos were caught by:

* codespell
* typos
* PyCharm

The code issues were identified by:

* Pycharm


actions_spec.openapi.yaml
```diff
@@ -138,7 +138,7 @@ paths:
           description: Authorization header missing or invalid.
         403:
           description: |
-            Not authorized. The requestor is not in the action's monitor_by
+            Not authorized. The requester is not in the action's monitor_by
```
sirosen, line 141: Not-a-problem, but this jumps out as not-really-a-fix in my mind. `requestor`, though less common, is usually accepted as an alternate spelling. So I hesitate to say that it's wrong -- at least, not in the way that `Exmaple` is wrong!

sirosen (approved):

actions_spec.openapi.yaml
kurtmckee: I thought the same thing, but Merriam-Webster's website couldn't find that spelling, even as a redirect, so I accepted it as a typo at face value.

PR #132, opened 2024-03-22
Support CORS requests to action introspect routes

kurtmckee:
Features
--------

*   Support CORS requests to introspection routes.

### DIFF
--- a/globus_action_provider_tools/flask/api_helpers.py
+++ b/globus_action_provider_tools/flask/api_helpers.py
@@ -205,7 +205,7 @@ def add_action_routes_to_blueprint(
             response.headers["Access-Control-Allow-Methods"] = "GET, OPTIONS"
             response.headers["Access-Control-Allow-Origin"] = "*"
             response.headers["Access-Control-Expose-Headers"] = "*"
-            return response, 200
+            return response, 204
 
         auth_state = check_token(request, checker)
         if not auth_state.check_authorization(
--- a/globus_action_provider_tools/flask/api_helpers.py
+++ b/globus_action_provider_tools/flask/api_helpers.py
@@ -197,8 +197,16 @@ def add_action_routes_to_blueprint(
     input_body_validator = get_input_body_validator(provider_description)
     blueprint.register_error_handler(Exception, blueprint_error_handler)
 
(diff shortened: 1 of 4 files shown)


globus_action_provider_tools/flask/api_helpers.py
```diff
@@ -197,8 +197,16 @@ def add_action_routes_to_blueprint(
...
-    @blueprint.route("/", methods=["GET"], strict_slashes=False)
+    @blueprint.route("/", methods=["GET", "OPTIONS"], strict_slashes=False)
     def action_introspect() -> ViewReturn:
+        # Short-circuit CORS requests.
+        if request.method == "OPTIONS":
+            response = flask.make_response("")
+            response.headers["Access-Control-Allow-Methods"] = "GET, OPTIONS"
+            response.headers["Access-Control-Allow-Origin"] = "*"
+            response.headers["Access-Control-Expose-Headers"] = "*"
+            return response, 200
```
jakeglobus, line 208: I think you want a `204` here, not a `200`. You can see an example response to the `OPTIONS` method [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) if that's helpful.

globus_action_provider_tools/flask/apt_blueprint.py
```diff
@@ -168,6 +169,15 @@ def _action_introspect(self):
...
         """
         self._register_route_type("introspect")
+
+        # Short-circuit CORS requests.
+        if request.method == "OPTIONS":
+            response = flask.make_response("")
+            response.headers["Access-Control-Allow-Methods"] = "GET, OPTIONS"
+            response.headers["Access-Control-Allow-Origin"] = "*"
+            response.headers["Access-Control-Expose-Headers"] = "*"
+            return response, 200
```
jakeglobus, line 179: Same here with the `204`

jakeglobus (commented):
I don't see `Access-Control-Allow-Origin: *` being returned on the `GET` part of your requests here, you will need that as well. 

Just to double check too, this isn't a request that requires credentials, right?

jakeglobus (approved):

PR #137, opened 2024-03-29
Reduce I/O with Globus Auth

kurtmckee:
Bugfixes
--------

-   Allow package consumers to run with Python optimizations enabled.

    This is supported by replacing ``assert`` statements with ``raise AssertionError``.

Changes
-------

-   Remove references to web browsers from HTTP 401 Unauthorized responses.
-   Reduce I/O with Globus Auth when possible.

    *   If the action provider is visible to ``"public"``,        introspection requests are allowed without checking tokens.
    *   If the bearer token is missing, malformed, or is too short or long,        the incoming request is summarily rejected with HTTP 401        without introspecting the token.

### DIFF
--- a/globus_action_provider_tools/flask/helpers.py
+++ b/globus_action_provider_tools/flask/helpers.py
@@ -250,7 +250,7 @@ def assign_json_provider(app_or_blueprint: flask.Flask | flask.Blueprint):
     """
 
     if json_provider_available:
-        assert JsonProvider is not None
+        assert JsonProvider is not None  # mypy hack
         app_or_blueprint.json = JsonProvider(app_or_blueprint)
     else:
         app_or_blueprint.json_encoder = ActionProviderJsonEncoder
--- a/globus_action_provider_tools/flask/helpers.py
+++ b/globus_action_provider_tools/flask/helpers.py
@@ -20,7 +20,10 @@ from globus_action_provider_tools.data_types import (
     RequestObject,
     convert_to_json,
 )
-from globus_action_provider_tools.errors import AuthenticationError
+from globus_action_provider_tools.errors import (
+    AuthenticationError,
+    UnverifiedAuthenticationError,
+)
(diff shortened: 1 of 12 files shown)


kurtmckee:
[sc-30785]

globus_action_provider_tools/flask/helpers.py
```diff
@@ -250,7 +257,7 @@ def assign_json_provider(app_or_blueprint: flask.Flask | flask.Blueprint):
     """
 
     if json_provider_available:
-        assert JsonProvider is not None
+        assert JsonProvider is not None  # mypy hack
```
sirosen, line 260: Basic thought about this, though it's OT for this PR:
The name is always defined. So the test above could be replaced with `if JsonProvider is not None` and everything would work out fine and the assert can go away.

---

Tastes vary, but I actually _don't_ consider the type-conforming asserts like this one to be hacks. IMO they're a perfectly reasonable use for asserts: safe to strip at runtime, adding minimal safety, low impact, and clear in meaning. (That first one being rather key.)
Given the 3 typical options for "explaining the situation to type checkers", we have "ignore", "cast", and "assert". But only assert has the effect of narrowing union types, rather than discarding the type information entirely.

I wish there were a "truly 0 runtime impact" version of the assert behavior. e.g. `# type: assert[x is not None]`. That would be great.

I feel like there's a sort of dated design thing here in that *probably* we were all meant to always run production code with `-O`, but obviously that didn't happen and now we're all stuck with "debug assertions which run even when you aren't debugging". But is it a "type checker hack", or is it "sadly, even though assertions are offered to us with the best of intentions, we need to treat them as always having runtime impact"?

... sorry for the long post. Just allowing myself to give voice to some "thoughts and feelings" about the unfortunate usability issues around type-narrowing assertions.

sirosen (approved):

PR #152, opened 2024-06-07
Eliminate `None` as a return type from authentication methods

kurtmckee:
This is related to the HTTP 429 Too Many Requests story. This is a small bit of disentanglement and simplification, and is scoped to be standalone.

> [!NOTE]
>
> A single test case changed, which might suggest a behavioral change. This is not the case.
>
> The `ValueError` that is now raised is already caught by an ancestor caller:
>
> ```
> WAS:
>
> .groups                                 (catch ValueError)
> `--- ._get_groups_client()              (if None, raise ValueError)
>      `--- .get_authorizer_for_scope()   (return None)
>
> NOW:
>
> .groups                                 (catch ValueError)
> `--- ._get_groups_client()
>      `--- .get_authorizer_for_scope()   (raise ValueError)
> ```
>
> `.groups` is the _ONLY_ caller for `.get_groups_client()`, and `.get_groups_client()` is the _ONLY_ caller of `.get_authorizer_for_scope()`.

Development
-----------

* Eliminate ``None`` as a return type from authentication-related methods.

### DIFF
--- a/globus_action_provider_tools/authentication.py
+++ b/globus_action_provider_tools/authentication.py
@@ -67,12 +67,7 @@ class AuthState:
         self.errors: list[Exception] = []
         self._groups_client: GroupsClient | None = None
 
-    def introspect_token(self) -> GlobusHTTPResponse | None:
-        # There are cases where a null or empty string bearer token are present as a
-        # placeholder
-        if self.bearer_token is None:
-            return None
-
+    def introspect_token(self) -> GlobusHTTPResponse:
         resp = AuthState.introspect_cache.get(self.bearer_token)
         if resp is not None:
             log.info(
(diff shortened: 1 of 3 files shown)


globus_action_provider_tools/authentication.py
```diff
@@ -84,51 +79,20 @@ def introspect_token(self) -> GlobusHTTPResponse | None:
         resp = self.auth_client.oauth2_token_introspect(
             self.bearer_token, include="identity_set"
         )
-        now = time()
 
-        try:
-            if resp.get("active", False) is not True:
```
sirosen, line 90: I didn't like the old way that this was written (with `assert`), and even after the improvements I think we can do better. But it looks like this change removes these checks without any other location which examines `"active"`. Is that correct?

To my understanding, `active` is a response key which we must examine to be oauth2 compliant. It's possible that we need some of these other checks as well -- I've been meaning to look into that -- but that's the one which I'm pretty sure we need.

Maybe these are checked elsewhere and I've forgotten. Just LMK if I should look around.

kurtmckee:
Closed in favor of #178.

PR #169, opened 2024-10-07
Replace 'TokenChecker' with 'AuthStateBuilder'

sirosen:
This change ends up touching many files as it is necessary to replace
'TokenChecker' wherever it appears. There are two primary drivers for
this change:

- move the confidential client construction *out of* the state
  builder, so that it is more visible and can be flagged as a location
  which needs attention in order for us to tune and control network
  behaviors

- remove the `client_name`/`expected_audience` from interfaces, where
  it cannot serve a useful purpose and will only confuse users and
  maintainers

It might be valuable to check the audience field in the future, to
guard against some flavor of "confused deputy" failures. However, such
faults are not realistically as much of a problem as simply *not
supplying the right data to your application* and the source for the
client name is typically going to just be the same as the source for
the credentials. A better protection would need ground-up design.

As a *beneficial* side-effect: remove the `testing` subpackage with no
planned replacement. These testing helpers do too little to be
generically useful and are a way in which the API surface of this
library is too large relative to the maintenance-hours we can give it.
It was necessary to either update these or delete them, which is why I
have chosen deletion at this point in time.

One unfortunate testing "casualty" of these changes is the testing of
the token inspection codepath. The situation we were in for that test
was unstable, and several changes here broke it -- some tests relied
upon the testing flask apps having fully mocked Auth information, but
that test relied upon the flask apps making *some* of their effectful
callouts. Rather than building a dedicated test app for that test, it
is simply moved to a more targetted unit test for now to simplify
maintenance of the testsuite.

---

Please specially note the `# FIXME: NETWORK-BEHAVIOR` comments.
These were the initial driver of this activity, but this "TokenChecker" misnaming has made it hard to read this code for far too long.


sirosen:
Upon review, I'm going to mark this as a draft. This change is my destination, but trying to go here in a single step is probably too much. I intend to rebase and un-draft after getting some preliminary changes in.

sirosen:
Replaced by #177

PR #170, opened 2024-10-08
Fix coverage reporting to count src/ properly

sirosen:


ada-globus (approved):
Approving this as the accepted incantation while explicitly not asserting understanding of what coverage is actually trying to do with this mapping.

PR #175, opened 2024-10-11
Refactor `get_authorizer_for_scope`

sirosen:
Better control where and how dependent token callouts happen, and note with inline FIXME comments several issues which persist even with the refined implementation.

Logically, there is almost no change, although the refactor may make it appear that there is one. `get_authorizer_for_scope` now retries by clearing the cache and fetching new dependent tokens on failure, and has clearer failure semantics.
This happened in the previous code as well -- if the access token is expired and the refresh token is missing, a second call to get dependent tokens is issued. However, due to an inaccurate type annotation (`refresh_token` is `cast(str, ...)`, where it should be `str | None`), it *appears* that there was a previously unreachable behavior -- a retry -- which is now reachable.

In practical fact, these subtleties do not yet make any difference, as `refresh_token` will always be present and will be a string if there is any data at all. Anything else is an invalid and unreachable state, given that refresh tokens are currently always requested.

With the refactor completed, we can tackle separate changes to actually alter behavior.

### DIFF
--- a/src/globus_action_provider_tools/authentication.py
+++ b/src/globus_action_provider_tools/authentication.py
@@ -4,7 +4,7 @@ import functools
 import hashlib
 import logging
 from time import time
-from typing import Iterable, cast
+from typing import Iterable
 
 import globus_sdk
 from cachetools import TTLCache
@@ -15,7 +15,6 @@ from globus_sdk import (
     GlobusError,
     GlobusHTTPResponse,
     GroupsClient,
-    OAuthTokenResponse,
     RefreshTokenAuthorizer,
 )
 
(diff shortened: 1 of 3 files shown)


src/globus_action_provider_tools/authentication.py
```diff
@@ -201,100 +207,134 @@ def get_dependent_tokens(self, bypass_cache_lookup=False) -> OAuthTokenResponse:
...
+
+        # FIXME: switch from dependent refresh tokens to dependent access tokens
+        #
+        # dependent refresh tokens in an ephemeral execution context are not appropriate
+        # because the dependent refresh tokens are cached in-memory, not in a persistent store of state,
+        # this means we're fetching long-lived credentials which are then not persisted into long term
+        # storage
+        #
+        # the discrepancy between the credential type and the storage strategy reveals that these tokens
+        # are only used in a synchronous context in which the refresh token is not needed or wanted
```
ada-globus, line 302: 💯

ada-globus (commented):
This looks good to me. Are there existing tests that exercise these functions? If not, is there anything we can add to run water through it, even if minimally to start? I have very little doubt about the correctness of the change, but want to make sure we've got at least basic coverage (even if we anticipate further changes).

sirosen:
I ended up on other adventures, but I returned to this and added a small suite of some basic tests.

PR #176, opened 2024-10-11
Make mypy stricter

sirosen:
- turn on a variety of mypy strictness flags, but stay short of
  `disallow_untyped_defs` -- which is why we aren't in strict mode
- `tox r -e mypy` does not pass any CLI flags
- cleanup numerous annotations
- in some cases, make minor implementation ordering adjustments to
  make things pass
- Put a small shim in place over TTLCache to allow its contents to be
  type checked

### DIFF
--- a/src/globus_action_provider_tools/utils.py
+++ b/src/globus_action_provider_tools/utils.py
@@ -11,6 +11,7 @@ class TypedTTLCache(t.Generic[T]):
     A tiny wrapper class which provides a type-checked layer on top of TTLCache.
     This allows us to know and enforce the types of cached objects.
     """
+
     def __init__(self, *, maxsize: int, ttl: int) -> None:
         self._cache: cachetools.TTLCache = cachetools.TTLCache(maxsize=maxsize, ttl=ttl)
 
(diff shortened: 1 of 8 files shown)


src/globus_action_provider_tools/utils.py
```diff
@@ -1,4 +1,34 @@
...
+        self._cache[key] = value
+
+    def __delitem__(self, key: str) -> None:
+        del self._cache[key]
+
+    def __getitem__(self, key: str) -> T:
+        return self._cache[key]  # type: ignore[no-any-return]
+
+    def __contains__(self, key: str) -> bool:
+        return key in self._cache
```
ada-globus, line 31: I think this needs:

```suggestion
    def __contains__(self, key: str) -> bool:
        return key in self._cache
    
    def clear(self) -> None:
        self.clear()
```

sirosen: _scratches head_
I swear I had passing CI. Maybe I had `clear()` and then I removed it by mistake? But yes, it needs all of the methods which we use. :grimacing:

ada-globus (approved):

PR #177, opened 2024-10-11
Replace 'TokenChecker' with 'AuthStateBuilder'

sirosen:
_Replaces #169. That changeset was `+195, -562`, this one, which relies on us having broken out the prereqs, ends up at a much more modest `+170, -115`. Hopefully this is now small enough that we can review and discuss._

This change replaces 'TokenChecker' wherever it appears.

This fixes some bad naming, but it also moves the confidential client construction *out of* the state builder, so that it is more visible and can examined as a location which needs attention in order for us to tune and control network behaviors.


sirosen:
100% agree. I just feel like that's a separate set of changes. :smile:

PR #178, opened 2024-10-18
Reduce the prevalence of `None` as a return value

kurtmckee:
_Supersedes #152_

Changes
-------

-   ``AuthState.introspect_token()`` will no longer return ``None`` if the token is not active.

    Instead, a new exception, ``InactiveTokenError``, will be raised.
    ``InactiveTokenError`` is a subclass of ``ValueError``.

    Existing code that calls ``AuthState.introspect_token()`` no longer returns ``None``, either,
    but will instead raise ``ValueError``, a subclass of ``ValueError``,
    or a ``globus_sdk.GlobusAPIError``:

    *   ``AuthState.get_authorizer_for_scope``
    *   ``AuthState.effective_identity``
    *   ``AuthState.identities``

### DIFF
--- a/changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
+++ b/changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
@@ -1,0 +1,15 @@
+Changes
+-------
+
+-   ``AuthState.introspect_token()`` will no longer return ``None`` if the token is not active.
+
+    Instead, a new exception, ``InactiveTokenError``, will be raised.
+    ``InactiveTokenError`` is a subclass of ``ValueError``.
+
+    Existing code that calls ``AuthState.introspect_token()`` no longer returns ``None``, either,
+    but will instead raise ``ValueError``, a subclass of ``ValueError``,
+    or a ``globus_sdk.GlobusAPIError``:
+
+    *   ``AuthState.get_authorizer_for_scope``
+    *   ``AuthState.effective_identity``
+    *   ``AuthState.identities``
--- a/changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
+++ b/changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
@@ -7,8 +7,7 @@ Changes
     ``InactiveTokenError`` is a subclass of ``ValueError``.
 
     Existing code that calls ``AuthState.introspect_token()`` no longer returns ``None``, either,
-    but will instead raise ``ValueError``, a subclass of ``ValueError``,
-    or a ``globus_sdk.GlobusAPIError``:
+    but will instead raise ``ValueError`` (or a subclass) or a ``globus_sdk.GlobusAPIError``:
 
     *   ``AuthState.get_authorizer_for_scope``
     *   ``AuthState.effective_identity``
--- a/src/globus_action_provider_tools/authentication.py
+++ b/src/globus_action_provider_tools/authentication.py
@@ -122,7 +122,7 @@ class AuthState:
         """
 
         if not introspect_result["active"]:
-            raise InactiveTokenError("The token is invalid or has been revoked")
+            raise InactiveTokenError("The token is invalid.")
 
         # validate scopes, ensuring that the token provided accords with the service's
         # notion of what operations exist and are supported
@@ -239,7 +239,7 @@ class AuthState:
         ``required_authorizer_expiration_time``, it is treated as a failure.
         """
 
-        had_cached_value, dependent_tokens = self._get_cached_dependent_tokens()
+        retrieved_from_cache, dependent_tokens = self._get_cached_dependent_tokens()
 
         # if the dependent token data (which could have been cached) failed to meet
         # the requirements...
@@ -251,8 +251,8 @@ class AuthState:
             # there's no reason to expect new tokens would do better
             # fail, but do not clear the cache -- it could be satisfactory
             # for some other scope request
-            if not had_cached_value:
-                raise ValueError("No matching cached dependent tokens found")
+            if not retrieved_from_cache:
+                raise ValueError("Dependent tokens do not match request.")
 
             # otherwise, the cached value was bad -- fetch and check again,
             # by clearing the cache and asking for the same data
@@ -264,7 +264,7 @@ class AuthState:
             if not self._dependent_token_response_satisfies_scope_request(
                 dependent_tokens, scope, required_authorizer_expiration_time
             ):
-                raise ValueError("Refreshed dependent tokens still don't match")
+                raise ValueError("Dependent tokens do not match request.")
 
         token_data = dependent_tokens.by_scopes[scope]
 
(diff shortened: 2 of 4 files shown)


src/globus_action_provider_tools/authentication.py
```diff
@@ -107,40 +111,34 @@ def _cached_introspect_call(self) -> GlobusHTTPResponse:
...
         self._verify_introspect_result(introspect_result)
         return introspect_result
 
     def _verify_introspect_result(self, introspect_result: GlobusHTTPResponse) -> None:
         """
         A helper which checks token introspect properties and raises exceptions on failure.
         """
+
+        if not introspect_result["active"]:
+            raise InactiveTokenError("The token is invalid or has been revoked")
```
sirosen, line 125:
```suggestion
            raise InactiveTokenError("The token is invalid.")
```

Small personal preference here.
Possible reasons for being invalid that I know of:
- the string is not an Auth-issued token string -- maybe it doesn't even parse internally, etc.
- the token expired
- the token was revoked explicitly by the user
- the token was not explicitly revoked, but our team found it in public and revoked it (e.g. we find a refresh token in someone's GitHub repo)

So I'd just say "it's invalid" and leave the possibility of revocation as something covered by that statement.

src/globus_action_provider_tools/authentication.py
```diff
@@ -240,39 +238,33 @@ def get_authorizer_for_scope(
...
+        if not self._dependent_token_response_satisfies_scope_request(
+            dependent_tokens, scope, required_authorizer_expiration_time
+        ):
+            # if there was no cached value, we just got fresh dependent tokens
+            # to do this work but they weren't sufficient
+            # there's no reason to expect new tokens would do better
+            # fail, but do not clear the cache -- it could be satisfactory
+            # for some other scope request
+            if not had_cached_value:
+                raise ValueError("No matching cached dependent tokens found")
```
sirosen, line 255: I'm reading the comment block above and feel like we can improve this message.
It feels like it's telling us "there was a cache miss", but the more complete message is
"there was a cache miss, so we _did_ do the callout, and it failed to get matching data"

I'm going to suggest we simply label these as the "cold" and "hot" cache paths, and trust that it provides enough information for us to find our way back to this context if necessary. 

So another suggestion follows this, but first:

```suggestion
                raise ValueError("Dependent tokens did not match request (cold cache)")
```

src/globus_action_provider_tools/authentication.py
```diff
@@ -240,39 +238,33 @@ def get_authorizer_for_scope(
...
-                    dependent_tokens, scope, required_authorizer_expiration_time
-                ):
-                    return None  # FIXME: raise an exception instead
-        except globus_sdk.AuthAPIError:
-            log.warning(
-                f"Unable to create GlobusAuthorizer for scope {scope}. Using 'None'",
-                exc_info=True,
-            )
-            return None
+                raise ValueError("Refreshed dependent tokens still don't match")
```
sirosen, line 267:
```suggestion
                raise ValueError("Dependent tokens did not match request (hot cache)")
```

sirosen (changes requested):
Some minor tweaks to apply, but I want to hold off on approving and merging until after we do a release.

Although we're now fairly certain that this won't cause impactful changes for known AP Tools users, I think including it in the next release would not match some of the messages I've sent to other teams about what to expect out of our next two releases.

changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
```diff
@@ -0,0 +1,15 @@
+Changes
+-------
+
+-   ``AuthState.introspect_token()`` will no longer return ``None`` if the token is not active.
+
+    Instead, a new exception, ``InactiveTokenError``, will be raised.
+    ``InactiveTokenError`` is a subclass of ``ValueError``.
+
+    Existing code that calls ``AuthState.introspect_token()`` no longer returns ``None``, either,
+    but will instead raise ``ValueError``, a subclass of ``ValueError``,
```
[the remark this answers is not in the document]
kurtmckee: `InactiveTokenError` and `InvalidTokenScopesError` are subclasses of `ValueError`...?

src/globus_action_provider_tools/authentication.py
kurtmckee: Done.

src/globus_action_provider_tools/authentication.py
kurtmckee: Done.

src/globus_action_provider_tools/authentication.py
kurtmckee: Done.

kurtmckee: Actually, I'm unresolving this. I don't think the terms "cold" and "hot" cache are helpful here, because in all failure cases we've submitted an API request to Auth.

I think these can both just say "Dependent tokens did not match request". The rest is simply just cache-busting details.

changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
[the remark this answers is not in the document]
sirosen: Right, but the text says that `ValueError` is a subclass of `ValueError`, unless I'm very confused. (And if I am, I'll count that as a request for a rephrase!)

src/globus_action_provider_tools/authentication.py
sirosen: Yeah, I'm down for that. No distinction between the two errors! :+1:

changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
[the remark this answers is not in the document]
sirosen: Oh! I see now. It's a list of options, of length 3,
- `ValueError`
- a subclass of `ValueError`
- `globus_sdk.GlobusAPIError`

I thought it was a list of length 2,
- `ValueError`, a subclass of `ValueError`
- `globus_sdk.GlobusAPIError`

Can we scratch "a subclass of `ValueError`" from the list, but now on different grounds, which is that it's unnecessary to specify?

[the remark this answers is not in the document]
kurtmckee: One last alternate:

> `ValueError` (or a subclass) or `globus_sdk.GlobusAPIError`

?

sirosen (approved):

changelog.d/20241018_082221_kurtmckee_eliminate_none_return_values.rst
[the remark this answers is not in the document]
sirosen: My approval should speak for itself.

But since it doesn't seem to speak up, "yes!" and "w00t!" and other exclamations of joy. I now understand the text and it says everything you wanted it to say!

sirosen (approved):

PR #181, opened 2024-10-21
hotfix/0.19.1

kurtmckee:
Bugfixes
--------

- When introspecting tokens, allow the introspected scopes to be a superset of required scopes.

  A bug in the scope comparison code flipped the logic;
  if a user consented to scopes A and B and the action provider required only scope A,
  the comparison would fail *as if A and B were required but only A had been consented to*.

  This is now fixed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,10 +10,28 @@ Unreleased changes are documented in files in the `changelog.d`_ directory.
 
 ..  scriv-insert-here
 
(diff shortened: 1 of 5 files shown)


sirosen (approved):

ada-globus (approved):

kurtmckee:
_Force-pushed to update the release date in the CHANGELOG_

PR #184, opened 2024-10-26
Fix erroneous use of dependent refresh tokens

sirosen:
The previous code (here, almost entirely removed) requested dependent refresh tokens unconditionally. These were then stored in a 47 hour cache, and there was never an update to the cache to refresh these credentials. Access tokens last for 48 hours, meaning that with the prior cache parameters, it was not realistically possible to reach the point of requiring a refresh via expiration. (A refresh triggered by token revocation would fail, and therefore is not worthy of special attention.)

Similar to the impossibility of a normal refresh, the timing requirement expressed by the `required_authorizer_expiration_time` parameter was more or less impossible to trigger in normal usage. With the default value of 60 seconds, the only way it could fire requires an hour to pass between the request being received and the dependent token for that request being retrieved. Therefore, this parameter is no longer checked, and emits a DeprecationWarning if used explicitly.

### DIFF
--- a/src/globus_action_provider_tools/authentication.py
+++ b/src/globus_action_provider_tools/authentication.py
@@ -219,7 +219,7 @@ class AuthState:
             warnings.warn(
                 "`required_authorizer_expiration_time` has no effect and will be removed in a future version.",
                 DeprecationWarning,
-                stacklevel=1,
+                stacklevel=2,
             )
 
         retrieved_from_cache, dependent_tokens = self._get_cached_dependent_tokens()
--- a/src/globus_action_provider_tools/authentication.py
+++ b/src/globus_action_provider_tools/authentication.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import functools
 import hashlib
 import logging
-import time
+import warnings
 from typing import Iterable
 
 import globus_sdk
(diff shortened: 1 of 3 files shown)


src/globus_action_provider_tools/authentication.py
```diff
@@ -218,39 +217,34 @@ def get_dependent_tokens(
...
-        If the access token is or will be expired within the
-        ``required_authorizer_expiration_time``, it is treated as a failure.
+        :param scope: The scope for which an authorizer is being requested
+        :param required_authorizer_expiration_time: Deprecated parameter. Has no effect.
         """
+        if required_authorizer_expiration_time is not None:
+            warnings.warn(
+                "`required_authorizer_expiration_time` has no effect and will be removed in a future version.",
+                DeprecationWarning,
+                stacklevel=1,
```
kurtmckee, line 236: I think this should be at least level 2 so people can see where in their code they've called this.

```suggestion
                stacklevel=2,
```

kurtmckee (changes requested):
I'd prefer to see tests that exercise the new `DeprecationWarning`.

src/globus_action_provider_tools/authentication.py
sirosen: Ah, yes, good catch!

I find this parameter hard to index correctly in my head because in some sense it counts from 1 rather than 0.

sirosen:
Test case for the deprecation added.

sirosen:
I rebased to clear the conflict -- it should build clean (passes for me locally), but let's watch CI.

kurtmckee (approved):

PR #187, opened 2024-11-04
Introduce 'ClientFactory' to tune transport-layer parameters attached to any of the client-building components

sirosen:
Introduce a ClientFactory pattern encoding params

A ClientFactory is a newly-defined interface specific to AP Tools which can construct ConfidentialAppAuthClient and GroupsClient instances with pre-baked initialization parameters.
In our case, the default implementation is customized to provide transport-layer parameters tuning the retry and timeout behaviors.

A user can implement the same interface and pass an alternative ClientFactory to the blueprint config, allowing them freedom to more tightly control the construction of Auth and Groups clients within AP Tools.

Minimal changes are made to add `client_factory` as an init-time argument for AuthStateBuilder and AuthState, and `client_factory` is added as an attribute of the blueprint config (ActionProviderConfig).
The blueprint will pass its configured factory to AuthStateBuilder, and the AuthStateBuilder will pass its configured factory to AuthState.

In order to keep tests fast, the ClientFactory is subclassed for testing, with implementations which set low-retry and 0-sleep behaviors via the supported params. Only one new test is introduced to validate the retry-once behavior of the default configuration.

### DIFF
--- a/changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
+++ b/changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
@@ -9,8 +9,8 @@ Features
 
   - The number of retries for both client types is reduced to 1 (from an
     SDK-default of 5).
-  - The http timeout is reduced to 30 seconds (from an SDK-default of 60s).
-  - The max sleep duration is reduced to 5 seconds (fram an SDK-default of
+  - The HTTP timeout is reduced to 30 seconds (from an SDK default of 60s).
+  - The max sleep duration is reduced to 5 seconds (from an SDK default of
     10s).
   - ActionProviderConfig, AuthStateBuilder, and AuthState are all customized to
     accept a ClientFactory, and to use the client factory for any client
--- a/changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
+++ b/changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
@@ -1,0 +1,17 @@
+Features
+--------
+
+- A new component, ``ClientFactory`` is now exposed in
+  ``globus_action_provider_tools.client_factory``. This allows users to
+  customize the transport-layer settings used for Auth and Groups clients which
+  are constructed by the Action Provider Tools library, and sets initial
+  parameters for this tuning.
+
+  - The number of retries for both client types is reduced to 1 (from an
+    SDK-default of 5).
+  - The http timeout is reduced to 30 seconds (from an SDK-default of 60s).
+  - The max sleep duration is reduced to 5 seconds (fram an SDK-default of
+    10s).
+  - ActionProviderConfig, AuthStateBuilder, and AuthState are all customized to
+    accept a ClientFactory, and to use the client factory for any client
+    building operations.
--- a/src/globus_action_provider_tools/authentication.py
+++ b/src/globus_action_provider_tools/authentication.py
@@ -13,6 +13,7 @@ from globus_sdk import (
     GlobusHTTPResponse,
 )
 
+from .client_factory import ClientFactory
 from .utils import TypedTTLCache
 
 log = logging.getLogger(__name__)
@@ -70,14 +71,17 @@ class AuthState:
         auth_client: ConfidentialAppAuthClient,
         bearer_token: str,
         expected_scopes: frozenset[str],
+        client_factory: ClientFactory | None = None,
     ) -> None:
         self.auth_client = auth_client
         self.bearer_token = bearer_token
         self.sanitized_token = self.bearer_token[-7:]
         self.expected_scopes = expected_scopes
+        self._client_factory = (
+            client_factory if client_factory is not None else ClientFactory()
+        )
 
         self.errors: list[Exception] = []
-        self._groups_client: globus_sdk.GroupsClient | None = None
 
     @functools.cached_property
     def _token_hash(self) -> str:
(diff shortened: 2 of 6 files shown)


changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
```diff
@@ -0,0 +1,17 @@
...
+
+- A new component, ``ClientFactory`` is now exposed in
+  ``globus_action_provider_tools.client_factory``. This allows users to
+  customize the transport-layer settings used for Auth and Groups clients which
+  are constructed by the Action Provider Tools library, and sets initial
+  parameters for this tuning.
+
+  - The number of retries for both client types is reduced to 1 (from an
+    SDK-default of 5).
+  - The http timeout is reduced to 30 seconds (from an SDK-default of 60s).
```
kurtmckee, line 12:
```suggestion
  - The HTP timeout is reduced to 30 seconds (from an SDK default of 60s).
```

changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
```diff
@@ -0,0 +1,17 @@
...
+  - The max sleep duration is reduced to 5 seconds (fram an SDK-default of
```
kurtmckee, line 13:
```suggestion
  - The max sleep duration is reduced to 5 seconds (from an SDK default of
```

src/globus_action_provider_tools/authentication.py
```diff
@@ -70,14 +71,17 @@ def __init__(
...
         expected_scopes: frozenset[str],
+        client_factory: ClientFactory | None = None,
     ) -> None:
         self.auth_client = auth_client
         self.bearer_token = bearer_token
         self.sanitized_token = self.bearer_token[-7:]
         self.expected_scopes = expected_scopes
+        self._client_factory = (
+            client_factory if client_factory is not None else ClientFactory()
+        )
```
kurtmckee, line 82: Is this equivalent?

```suggestion
        self._client_factory = client_factory or ClientFactory()
```

src/globus_action_provider_tools/authentication.py
```diff
@@ -317,9 +331,12 @@ def __init__(
         self,
         auth_client: globus_sdk.ConfidentialAppAuthClient,
         expected_scopes: Iterable[str],
+        *,
+        client_factory: ClientFactory | None = None,
     ) -> None:
         self.auth_client = auth_client
         self.default_expected_scopes = frozenset(expected_scopes)
+        self.client_factory = client_factory
```
kurtmckee, line 339: Does this need to instantiate `ClientFactory` if `client_factory` is `None`?

src/globus_action_provider_tools/client_factory.py
```diff
@@ -0,0 +1,49 @@
...
+        ("max_sleep", 5),
+    )
+
+    def make_confidential_app_auth_client(
+        self, client_id: str | uuid.UUID, client_secret: str
+    ) -> globus_sdk.ConfidentialAppAuthClient:
+        return globus_sdk.ConfidentialAppAuthClient(
+            client_id=client_id,
+            client_secret=client_secret,
+            transport_params={k: v for k, v in self.DEFAULT_AUTH_TRANSPORT_PARAMS},
```
kurtmckee, line 37:
```suggestion
            transport_params=dict(self.DEFAULT_AUTH_TRANSPORT_PARAMS),
```

src/globus_action_provider_tools/client_factory.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def make_groups_client(
+        self,
+        authorizer: (
+            globus_sdk.AccessTokenAuthorizer | globus_sdk.RefreshTokenAuthorizer | None
+        ),
+    ) -> globus_sdk.GroupsClient:
+        return globus_sdk.GroupsClient(
+            authorizer=authorizer,
+            transport_params={k: v for k, v in self.DEFAULT_GROUPS_TRANSPORT_PARAMS},
```
kurtmckee, line 48:
```suggestion
            transport_params=dict(self.DEFAULT_GROUPS_TRANSPORT_PARAMS),
```

tests/test_authentication.py
```diff
@@ -118,15 +142,58 @@ def test_invalid_grant_exception(auth_state):
...
+
+    On two 5xx responses followed by a 200, getting an authorizer fails.
+    (Default configuration)
+    """
+    auth_state = get_auth_state_instance(
+        ["expected-scope"], client_factory=FastRetryClientFactory()
+    )
+    body_for_200s = [
+        {
+            "access_token": "oompa-loompa-doompa-de-access-token",
```
kurtmckee, line 169: This makes me so happy. :heart:

tests/test_authentication.py
```diff
@@ -118,15 +142,58 @@ def test_invalid_grant_exception(auth_state):
...
+            method="POST",
+            status=status,
+            json=body,
+        ).add()
+
+    if succeeds:
+        authorizer = auth_state.get_authorizer_for_scope("golden-ticket")
+        assert isinstance(authorizer, globus_sdk.AccessTokenAuthorizer)
+    else:
+        with pytest.raises(globus_sdk.GlobusAPIError):
```
kurtmckee, line 195: I recommend checking what's in the exception (a message, a value, a shape of data, etc).

kurtmckee (changes requested):

src/globus_action_provider_tools/authentication.py
sirosen: It was intentionally omitted since it's not strictly necessary to create it here (since it gets handled in `AuthState.__init__`), but now that you've asked, I see two reasons to do it here:
- more efficient (instantiate once rather than N times)
- narrows `AuthStateBuilder.client_factory` on instances from `ClientFactory | None` to `ClientFactory`

So I'll change that.

src/globus_action_provider_tools/authentication.py
sirosen: Yeah, I'm applying the change you suggested.

I've gotten bitten by truthiness checks vs `is None` in the past, and I still don't always trivially decide whether to use the implicit `bool()` logic or `is None` checking. I think one rule, which this case matches, is easy: it's fine to use implicit `bool()` when the non-None type always `bool()`s as `True`.

changelog.d/20241104_170625_sirosen_client_factory_sets_params.rst
sirosen: "fram" :grimacing:

src/globus_action_provider_tools/client_factory.py
sirosen: Not sure why I couldn't see this! Good call.

kurtmckee (approved):

PR #189, opened 2024-11-06
Migrate to the reusable tox workflow

kurtmckee:
This PR migrates the project to the reusable tox workflow.

| Metric | Before this PR | This PR (cache miss) | This PR (cache hit) |
|--------|--------|--------|--------|
| Total duration | [1m 30s](https://github.com/globus/action-provider-tools/actions/runs/11709001281) | [1m 15s](https://github.com/globus/action-provider-tools/actions/runs/11709054104?pr=189) | [44s](https://github.com/globus/action-provider-tools/actions/runs/11709095355?pr=189) |
| Total run time | [5m 41s](https://github.com/globus/action-provider-tools/actions/runs/11709001281/usage) | [2m 48s](https://github.com/globus/action-provider-tools/actions/runs/11709054104/usage?pr=189) | [1m 33s](https://github.com/globus/action-provider-tools/actions/runs/11709095355/usage?pr=189) |


sirosen (approved):
It looks like the prior logic would do a single wheel build which was then shared across jobs. That would be an interesting feature to support in the future.

kurtmckee:
That was an older technique I discovered on my quest for speed.

It was superseded once I found out how to configure a shared wheel across all tox environments, which is significantly faster.

sirosen:
Is there no longer any role for it?
I was considering the potential benefit of building a wheel and sharing it across parallel jobs. But I suppose that needs the tox jobs to accept that wheel (rather than rebuilding it themselves).

kurtmckee:
In terms of speed, no, I don't think there's a significant advantage, unless there are more runners spinning up than this.

The cost of spinning up and setting up a Linux runner to build a wheel has to be divided by the number of jobs that are dependent on that wheel (so, ~20/4 at the fastest), for a cost of ~5 seconds that would have to be saved per dependent job.

Back when I didn't know I could build the wheel once for all of the tox environments, there was a cost multiplier based on the number of tox environments. 8 tox environments on Linux meant ~16 seconds of wheel building. 2 Pythons on Windows might each take ~5 seconds each to build the wheel, for a cost of ~10 seconds.

Now that a wheel gets built once, and I'm executing so many tox environments per runner, the total cost of building a wheel per runner is so low that it's not justifiable to spin up a runner just to build the wheel.

PR #190, opened 2024-11-07
Update scriv categories

kurtmckee:
Development
-----------

*   Introduce new scriv categories to better communicate how the project evolves.

    The categories are also re-ordered,
    which defines how fragments will be ordered in the CHANGELOG.

*   Add a changelog fragment template.


ada-globus (approved):
Love this!

PR #193, opened 2024-11-18
Fix theme issues and revise incorrect intro copy

ada-globus:
I didn't attempt to fully remedy all that probably could/should be; I just made the minimum viable set of moves to address everything in my immediate field of view.


docs/source/conf.py
```diff
@@ -42,32 +42,33 @@
...
+html_show_sourcelink = True
 
 html_sidebars = {
-    "**": ["logo-text.html", "globaltoc.html", "localtoc.html", "searchbox.html"]
+    "**": [
+        "sidebar/scroll-start.html",
+        "sidebar/brand.html",
+        "sidebar/search.html",
+        "sidebar/navigation.html",
+        "sidebar/ethical-ads.html",
```
kurtmckee, line 69: Hang on, what's this bit right here?

ada-globus: Ahh, I think we can omit that, actually—I suspect the default is fine for our purpose. I'll strike it.

kurtmckee: Does the default include that `ethical-ads` bit?

kurtmckee (approved):

PR #201, opened 2024-12-05
Update tests to actually treat 'flask' as optional

sirosen:
1. Add a tox factor, `flask`, for installing flask via the `flask` extra
2. Update the tox matrix to run with and without `-flask` in addition to `-minimum_flask`
3. Make test collection ignore the `flask/` test dir if `flask` can't be imported (also, rename the matching tests to `tests/flask/`)
4. Minor changes in those tests so that the `conftest.py` won't fail during test collection

---

There's a long-standing inconsistency in that the tests have always required `flask` even though the library declares it to be an optional dependency.
We may soon have an actual usage of this library (an AP CLI client) which doesn't need `flask`, so resolving this one way or the other will help ensure that installs don't have broken imports if `flask` is missing.


tests/conftest.py
```diff
@@ -16,6 +16,13 @@
 
 from .data import canned_responses
 
+try:
+    import flask
+
+    assert flask
```
kurtmckee, line 22: What does `assert flask` do?

kurtmckee (approved):
I think I was expecting some tests to be marked with a `skipif` or `skipunlessimport` (or whatever it's called), or to have a custom pytest marker associated with the Flask tests.

Is that what the `collect_ignore` in `conftest.py` is doing? If so, then I've learned a very neat new thing!

tests/conftest.py
sirosen: Oh, I did that on the fly for flake8's F401. Let me replace with a noqa comment.

sirosen:
Yeah, I tried out the `collect_ignore` instead of a skip marker or similar because I wanted to avoid even trying to load those modules (and dealing with their imports). It's new to me, and we can obviously always switch to a skip marker if we find it doesn't work well.

One interesting thing is that even with that collect option set, the `conftest.py` in the relevant subdirectory gets loaded. To me, that was a big surprise! So I wouldn't call this a "new and recommended pattern" yet -- just a different way of doing things which seemed to fit this use case nicely.

kurtmckee (approved):

PR #203, opened 2025-02-01
Rewrite the `setting_up_auth` document

sirosen:
Every line of the doc has been evaluated and for the most part
replaced. The doc is more prescriptive and based around use of
`globus_sdk` in order to provide a good and ergonomic developer
experience.

Concepts are explained *minimally* but the focus is on getting the
client created and its scope created. Because this document does not
cover actually configuring an AP using the library, no mention is made
whatsoever of the `globus_auth_scope` field. However, we do note that
there is exactly one scope per AP. (Although not technically a strict
requirement, it is a practical fact, and mixing scopes from multiple
resource servers would cause failures.)

In the course of the revised guide, we walk the user through
constructing a `manage-ap.py` script which can act as their admin
interface for their AP. This will let them manage their scope as their
application evolves.

Following a best practice from the broader Python community, this doc
introduces some tabbed layout (using `sphinx-design`) in order to
provide instructions using the `py` runner for environment creation on
Windows. (See the `pip` documentation for other examples of this
practice.) In order to add `sphinx-design`, I have updated the nested
'docs' project and used `uvx --from 'poetry==1.8.4' poetry ...` to get
semantics which exactly match the version which was used when those
files were created and minimize the changes introduced here.


docs/source/setting_up_auth.rst
```diff
@@ -3,293 +3,327 @@
...
-authentication via the `Globus Auth
-<https://docs.globus.org/api/auth/specification/>`_ system. To
-authenticate RESTful requests using Globus Auth, a service must register as a
-"resource server". This is a multi-step process involving use of both the Globus
-Auth developer portal and the Globus Auth API for configuring various access
-control states. To help with this process, we provide a step-by-step guide to
-using Globus Auth for this purpose
+In the Globus ecosystem, services use
+`Globus Auth <https://docs.globus.org/api/auth/>`_
+to handle user authentication. In order for an Action Provider to function, it
```
ada-globus, line 8: I'd prefer we don't use title case for terminology (e.g., "Action Provider") because it puts us in situations where, for instance, we say "Flows" and "Flow" in the same doc, which I think actually harm understanding more than helping (not specifically the case in this doc—but this is why we've moved away elsewhere). I'm okay with an alternate strategy, though. I think in the docs site we've moved to bolding them. 

```suggestion
to handle user authentication. In order for an **action provider** to function, it
```

docs/source/setting_up_auth.rst
```diff
@@ -3,293 +3,327 @@
...
+needs to be configured as a service in Globus Auth.
 
-.. note::
-    In the examples below, we will use the command line tool ``curl`` to
-    perform the HTTP operations as it is widely available. We also use the
-    command line tool ``jq`` to format the ``curl`` command's json responses.
-    However, other tools and clients exist for interacting with REST and HTTP
-    services, so you may need to translate the ``curl`` and ``jq`` commands to
-    your preferred tools.
+This guide doc will walk you through the setup process.
```
ada-globus, line 11:
```suggestion
This guide doc will walk you through the setup process.

As part of this process, we will build a small Python application to manage your
Globus Auth scopes. Managing your application's resources via code is a best
practice and will make your action provider easier to build and maintain.
```

sirosen: Yep; good addition. Accepted.

docs/source/setting_up_auth.rst
```diff
@@ -3,293 +3,327 @@
...
-    document, returned on the Action Provider Introspection operation (``GET
-    /``) with the key ``globus_auth_scope``.
+At this stage, you have a scope for your Action Provider!
 
-- | Verify that the created scope(s) are correctly associated with the Action
-    Provider:
+You can think of the scope under two identifiers:
 
-    .. code-block:: BASH
+- the full string: ``"https://auth.globus.org/scopes/$CLIENT_ID/action_all"``
```
ada-globus, line 281:
```suggestion
- the full scope string: ``"https://auth.globus.org/scopes/$CLIENT_ID/action_all"``
```

docs/source/setting_up_auth.rst
```diff
@@ -3,293 +3,327 @@
...
-            https://auth.globus.org/v2/api/clients/$CLIENT_ID | jq
+The full string is globally unique. Even if another application registers
+``action_all``, it won't conflict with your application's scope.
+The suffix is only unique to your application.
 
-- | Once your app and its scope(s) have been created and verified, remove your
-    credentials from your command line environment. Be sure to take note of the
-    client ID and its associated client secret for use in other places in the toolkit.
+For this reason, when communicating with other services you will always use
+the full string.
```
ada-globus, line 289:
```suggestion
the full scope string.
```

ada-globus (approved):

PR #205, opened 2025-02-07
Eagerly check token validity and handle invalid tokens in flask helpers

sirosen:
This changeset makes one primary breaking change and calls out another (more subtle) one as well.

The first and most notable change here is that `AuthState` now calls `AuthState.introspect_token()` on init.
This removes error behaviors which were previously lazy -- for example, the first access to `AuthState.effective_identity` could raise an error.

Instead, users of `AuthState` should anticipate that the token has already been validated.
This does not guarantee that no calls to the `AuthState` will do IO and result in errors -- notably the codepath which reaches out to Groups is intentionally left untouched here for two reasons:

1. To limit the scope and scale of this change (and release)
2. To ensure that services which have no need for Groups information are not calling out to Groups (current supposition is that almost no AP usages actually interrogate Groups and we do not wish to change that)

Secondarily, now that `AuthState(...)` can raise an error, it is clearly possible to handle that gracefully in the `FlaskAuthStateBuilder`.
That component now translates the generic errors which are used by `AuthState` into an appropriate error for the flask components, which then gets handled and translated into an `Unauthorized` error in the default handler.
Because `FlaskAuthStateBuilder` has new and different error raising behavior, it is also noted as a breaking change.

---

In the course of testing this, I also addressed some unnecessarily aggressive mocking in the testsuite, which resulted in a number of tests evaluating against mocks when it is perfectly feasible for them to evaluate against the real `AuthState` object.

---

_NB: I assessed some other paths to resolving the issues with uncaught errors, like modifying the flask-blueprint error handler more extensively.
Although many such options are reasonable, my determination was that they do not solve the underlying issue, which is the ability of this object to make errors appear on property access at unpredictable times in the application lifecycle._

### DIFF
--- a/src/globus_action_provider_tools/flask/helpers.py
+++ b/src/globus_action_provider_tools/flask/helpers.py
@@ -82,9 +82,9 @@ class FlaskAuthStateBuilder(AuthStateBuilder):
         try:
             return super().build(access_token)
         except InvalidTokenScopesError as err:
-            raise AuthenticationError("token had invalid scopes") from err
+            raise AuthenticationError("Token has invalid scopes") from err
         except InactiveTokenError as err:
-            raise AuthenticationError("token was invalid, expired, or revoked") from err
+            raise AuthenticationError("Token is invalid, expired, or revoked") from err
 
 
 def parse_query_args(
--- a/src/globus_action_provider_tools/flask/helpers.py
+++ b/src/globus_action_provider_tools/flask/helpers.py
@@ -12,7 +12,12 @@ import jsonschema
 from flask import Request, current_app, jsonify
 from pydantic import BaseModel, ValidationError
 
-from globus_action_provider_tools.authentication import AuthState, AuthStateBuilder
+from globus_action_provider_tools.authentication import (
+    AuthState,
+    AuthStateBuilder,
+    InactiveTokenError,
+    InvalidTokenScopesError,
+)
 from globus_action_provider_tools.data_types import (
     ActionProviderDescription,
     ActionProviderJsonEncoder,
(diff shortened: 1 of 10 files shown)


src/globus_action_provider_tools/flask/helpers.py
```diff
@@ -74,7 +79,12 @@ def build_from_request(self, *, request: Request | None = None) -> AuthState:
         if not 10 <= len(access_token) <= 2048:
             raise UnverifiedAuthenticationError("Bearer token length is unexpected")
 
-        return super().build(access_token)
+        try:
+            return super().build(access_token)
+        except InvalidTokenScopesError as err:
+            raise AuthenticationError("token had invalid scopes") from err
+        except InactiveTokenError as err:
+            raise AuthenticationError("token was invalid, expired, or revoked") from err
```
kurtmckee, line 87: Past tense feels incorrect here.

Also, I prefer that our exception text follow sentence capitalization (I'm flexible on closing punctuation). All existing exceptions raised from this method are sentence-capitalized, so these additions clearly flag different authorship.

```suggestion
            raise AuthenticationError("Token has invalid scopes") from err
        except InactiveTokenError as err:
            raise AuthenticationError("Token is invalid, expired, or revoked") from err
```

kurtmckee (changes requested):

src/globus_action_provider_tools/flask/helpers.py
sirosen: Yep, I also prefer title case. As evidenced here, I'm not perfectly consistent in my application of that rule.

As for past vs present... I can't say that I find either one better than the other, so I'm happy to apply the suggestion!
(I read the choices as the near equivalent: "It is invalid now, at the time of my evaluation" vs "It was invalid a moment ago, when I evaluated it")

